PR #6, opened 2020-06-16
Use npm instead of yarn

jafeltra:
Small changes to make using `npm` our default instead of using `yarn` including:
- replaced yarn commands with npm ones
- removed yarn.lock
- added package-lock


SamCitron:
Looks good to me!

PR #9, opened 2020-07-15
Updated React app rewired

SamCitron:
React app now runs sushi and is no longer ejected. Small changes to App.js and RunButton.js as well as small changes to the RunButton.test.js


SamCitron:
This pull request has all changes needed for SUSHI to run in the playground. It is ready for review.

SamCitron:
@mint-thompson Your comments on the last SUSHI PR for this issue have been addressed in these changes.

src/tests/utils/Processing.test.js
```diff
@@ -0,0 +1,88 @@
...
+      });
+      const loadIntoPlaygroundSpy = jest.spyOn(loadModule, 'loadIntoDefsPlayground').mockImplementation(() => {
+        return undefined;
+      });
+      const dependencyDefs = loadExternalDependenciesPlayground(defs, version);
+      await dependencyDefs;
+      expect(unzipSpy).toBeCalled();
+      expect(loadDefsSpy).toBeCalled();
+      expect(loadIntoPlaygroundSpy).toBeCalled();
+    }, 99999999);
```
jafeltra, line 52: Is the timeout here and on line 74 still needed? I remember we were messing around with different async and await things, but we might have figured all that out and the timeout `99999999` might not be needed anymore!

src/utils/PlaygroundApp.js
```diff
@@ -0,0 +1,132 @@
...
+import { Configuration } from 'fsh-sushi/dist/fshtypes';
+import { loadExternalDependenciesPlayground, fillTank, readConfigPlayground } from './Processing';
+
+export async function playgroundApp(input) {
+  // Hard Code config
+  let config = Configuration;
+  try {
+    config = readConfigPlayground();
+  } catch {
+    console.log('Something went wrong when creating the configuration');
```
jafeltra, line 14: I think for consistency, we might want to use the `logger` here and in the few other `console.log`s you added here! I think in the cases that we are logging messages about errors, you can use `logger.error`.

src/tests/utils/Processing.test.js
```diff
@@ -0,0 +1,88 @@
...
+    it('should hard code the preferred config', () => {
+      const config = readConfigPlayground();
+      expect(config.fhirVersion).toEqual(['4.0.1']);
+      expect(config.id).toEqual('fhir.us.minimal');
+      expect(config.FSHOnly).toBe(false);
+    });
+  });
+
+  describe('#loadExternalDependenciesPlayground()', () => {
+    beforeAll((done) => {
```
jafeltra, line 21: I'm not sure I've used this before. Can you explain what the `beforeAll` and `afterAll` are doing here so I know once you leave?

src/tests/utils/Processing.test.js
```diff
@@ -0,0 +1,88 @@
+import temp from 'temp';
+import { loadExternalDependenciesPlayground, readConfigPlayground, fillTank } from '../../utils/Processing';
+import { RawFSH } from 'fsh-sushi/dist/import/RawFSH';
+import { FHIRDefinitions } from 'fsh-sushi/dist/fhirdefs';
+import * as loadModule from '../../utils/Load';
+import 'fake-indexeddb/auto';
+
+describe('Processing', () => {
+  temp.track();
```
jafeltra, line 9: I think this might have been left over from SUSHI, which tracks some temporary folders. Do you know if it's needed here?

src/tests/utils/Processing.test.js
SamCitron: This, like the manual long timeout on the end of the tests were leftover attempts at getting the tests to not timeout. (It didn't end up working) I removed it!

package.json
```diff
@@ -10,19 +10,23 @@
...
     "codemirror": "^5.54.0",
+    "fsh-sushi": "^0.14.0",
+    "null-loader": "^4.0.0",
     "react": "^16.13.1",
     "react-codemirror2": "^7.2.1",
     "react-dom": "^16.13.1",
-    "react-scripts": "3.4.1"
+    "react-scripts": "3.4.1",
+    "tar-stream": "^2.1.3",
+    "temp": "^0.9.1"
```
jafeltra, line 20: I think you can do an `npm uninstall temp` since you removed the only place it was used!

src/utils/PlaygroundApp.js
```diff
@@ -0,0 +1,47 @@
+import { FSHTank, RawFSH } from 'fsh-sushi/dist/import';
+import { exportFHIR } from 'fsh-sushi/dist/export';
+import { logger, Type } from 'fsh-sushi/dist/utils';
+import { FHIRDefinitions } from 'fsh-sushi/dist/fhirdefs';
+import { Configuration } from 'fsh-sushi/dist/fshtypes';
```
jafeltra, line 5: I'm not entirely sure why the linter didn't pick this one up, but I don't think `Configuration` is ever used, so I don't think you need this import.

src/tests/utils/PlaygroundApp.test.js
```diff
@@ -0,0 +1,51 @@
...
+    const loadSpyConfig = jest
+      .spyOn(processing, 'loadExternalDependenciesPlayground')
+      .mockReset()
+      .mockResolvedValue(new FHIRDefinitions());
+
+    const outPackage = await playgroundApp();
+    expect(configSpy).toHaveBeenCalled();
+    expect(loadSpyConfig).toHaveBeenCalledTimes(0);
+    expect(outPackage).toBeUndefined();
+  });
```
jafeltra, line 50: If this test is really difficult, we can probably live without it for now. I think the only case you don't have a test here for is the case if the `fillTank` throws an error. I think maybe we should have a test similar to the one for the config that checks the `exportFHIR` function is never called and we return undefined if `fillTank` throws an error.

src/tests/utils/Processing.test.js
jafeltra: Ah that makes sense! Thanks!

src/utils/PlaygroundApp.js
jafeltra: Oh wait I just noticed that we do set `let config = Configuration` below, but I don't think that's exactly right. `Configuration` is a type in SUSHI, and we can't really set a variable to be the type, so I don't think the initial setting of `config` is necessary and we can just wait for it to be set when we run `readConfigPlayground` and then the import isn't needed.

jafeltra:
One other thought I had was around the names of these functions. I think we had appended "Playground" to most of the functions when there was a load for normal SUSHi and a load for the playground all in one code base. Now that it's broken out, what do you think about potentially removing the "Playground" at the end? Or does it still make sense to make it clear these are distinct functions for the playground? I'm definitely open to thoughts on this one!

src/tests/utils/PlaygroundApp.test.js
jafeltra: Looks good! Thanks for adding it!

src/utils/RunSUSHI.js
```diff
@@ -0,0 +1,46 @@
...
+
+export async function runSUSHI(input) {
+  // Hard Code config
+  let config = null;
+  try {
+    config = readConfig();
+  } catch {
+    logger.error('Something went wrong when creating the configuration');
+    return;
+  }
```
ngfreiter, line 15: I think we can actually get away with a more lightweight approach here. Most of the config is actually not necessary for turning FSH -> JSON, which is what the Playground is doing. Most of the config is for the IG Publisher, but the Playground (at least at this point) does not care about the IG Publisher. The only property that should really be strictly necessary is `canonical`. So I think you can replace all this code with just:
```
const config = { canonical: "http://default.org" }
```
So `minimalConfig` could go, the `readConfig` function could go, and SUSHI should be able to run export to JSON just fine with only the `canonical`. I'm almost definitely the goof who told you to use `minimalConfig` in the first place, so sorry about that.

package.json
jafeltra: I just pushed some changes I got to package-lock.json, I'm not sure why I got them and you didn't!

src/utils/RunSUSHI.js
SamCitron: Sounds good that made things way easier.

src/tests/components/RunButton.test.js
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
 import { render } from '@testing-library/react';
 import { unmountComponentAtNode } from 'react-dom';
-import RunButton from '../components/RunButton';
+import RunButton from '../../components/RunButton';
 import { act } from 'react-dom/test-utils';
+import 'fake-indexeddb/auto';
```
ngfreiter, line 6: Do we need `fake-indexeddb` in this file? I am suspicious that we do not.

src/tests/utils/RunSUSHI.test.js
```diff
@@ -0,0 +1,50 @@
...
+    const outPackage = await runSUSHI(text);
+    expect(loadSpy).toHaveBeenCalled();
+    expect(outPackage.profiles).toHaveLength(1);
+  });
+
+  it('should return an empty package when fillTank does not execute properly', async () => {
+    const FHIRDefs = new FHIRDefinitions();
+    FHIRDefs.add(Patient);
+    FHIRDefs.add(StructureDefinition);
+    // const configSpy = jest.spyOn(processing, 'readConfig').mockReset().mockReturnValue(undefined);
```
ngfreiter, line 35: I think we can now get rid of this line.

src/tests/utils/RunSUSHI.test.js
```diff
@@ -0,0 +1,50 @@
...
+    const loadSpy = jest.spyOn(processing, 'loadExternalDependencies').mockReset().mockResolvedValue(FHIRDefs);
+    const fillTankSpy = jest
+      .spyOn(processing, 'fillTank')
+      .mockReset()
+      .mockImplementation(() => {
+        throw new Error('Failed to fill tank');
+      });
+    const input = 'Improper FSH code!';
+    const outPackage = await runSUSHI(input);
+    //expect(configSpy).toHaveBeenCalled();
```
ngfreiter, line 45: And this line as well.

src/tests/components/RunButton.test.js
SamCitron: I was also suspicious but any call to RunSUSHI won't go through in a testing environment without it.

ngfreiter: I'm fine with just ignoring my suspicions then!

PR #10, opened 2020-07-21
Display JSON

SamCitron:
Fixes CIMPL-407 linked [here](https://standardhealthrecord.atlassian.net/secure/RapidBoard.jspa?rapidView=9&projectKey=INT&modal=detail&selectedIssue=CIMPL-407&assignee=5ece9e7747d31e0c2a0e7681). Displays JSON output on the right side upon push of the run button. Doesn't display the config on purpose, however, the JSON output as of right now is not formatted very well for readability.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -23,29 +23,32 @@ export default function App() {
   const classes = useStyles();
 
   const [shouldRunSUSHI, setShouldRunSUSHI] = useState(false);
-  const [text, setText] = useState('Edit FSH Here!');
+  const [inputText, setInputText] = useState('Edit FSH Here!');
+  const [outputText, setOutputText] = useState('Your JSON Output Will Display Here: ');
 
-  function updateShouldRunSUSHI(shouldRunSUSHI) {
+  function buttonFunctionality(shouldRunSUSHI, sushiOutput) {
     setShouldRunSUSHI(shouldRunSUSHI);
+    setOutputText(sushiOutput);
+    // console.log(sushiOutput);
   }
-  function updateTextValue(text) {
-    setText(text);
+  function updateInputTextValue(text) {
+    setInputText(text);
   }
 
   return (
     <div className="root">
       <TopBar />
-      <RunButton onClick={updateShouldRunSUSHI} text={text} />
+      <RunButton onClick={buttonFunctionality} text={inputText} />
       <Grid className={classes.container} container>
         <Grid className={classes.itemTop} item xs={6}>
           <CodeMirrorComponent
-            value={text}
-            updateTextValue={updateTextValue}
-            updateShouldRunSUSHI={updateShouldRunSUSHI}
+            value={inputText}
+            updateTextValue={updateInputTextValue}
+            updateShouldRunSUSHI={buttonFunctionality}
           />
         </Grid>
         <Grid className={classes.itemTop} item xs={6}>
-          <JSONOutput shouldDisplaySUSHI={shouldRunSUSHI} text={text} />
+          <JSONOutput shouldDisplaySUSHI={shouldRunSUSHI} text={outputText} />
         </Grid>
         <Grid className={classes.itemBottom} item xs={12}>
           <ConsoleComponent />
--- a/src/App.js
+++ b/src/App.js
@@ -26,7 +26,7 @@ export default function App() {
   const [inputText, setInputText] = useState('Edit FSH Here!');
   const [outputText, setOutputText] = useState('Your JSON Output Will Display Here: ');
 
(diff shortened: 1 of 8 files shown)


src/components/JSONOutput.js
```diff
@@ -18,15 +20,15 @@ export default function JSONOutput(props) {
   //Checks to insure the shouldRunSUSHI is true (aka button has been pressed) and there is text to display for the output
   if (props.shouldDisplaySUSHI && props.text) {
     return (
-      <Box className={classes.box} border={1}>
-        <h4>Your Output:</h4>
+      <Box className={classes.box} border={1} overflow="scroll">
+        <h4>Your Output: </h4>
         <p>{props.text}</p>
```
jafeltra, line 25: @cmoesel has taught me the wonders of `<pre>`. If you replace this `<p>` tag with `<pre>` the JSON will be spaced out with the `\t` you gave it in `JSON.stringify()`.

src/components/RunButton.js
```diff
@@ -20,14 +20,35 @@ const useStyles = makeStyles((theme) => ({
...
+
 export default function RunButton(props) {
   const classes = useStyles();
 
   //Sets the shouldRunSUSHI to true
   async function handleClick() {
-    props.onClick(true);
     const outPackage = await runSUSHI(props.text);
-    console.log(outPackage);
+    let jsonOutput = JSON.stringify(outPackage, replacer, '\t');
```
jafeltra, line 36: If you do end up using the `<pre>` tag, the tabs look pretty indented. I often use 2 spaces to indent JSON, so you can replace the `'\t'` here with `2`.

src/App.js
```diff
@@ -23,29 +23,31 @@ export default function App() {
   const classes = useStyles();
 
   const [shouldRunSUSHI, setShouldRunSUSHI] = useState(false);
-  const [text, setText] = useState('Edit FSH Here!');
+  const [inputText, setInputText] = useState('Edit FSH Here!');
+  const [outputText, setOutputText] = useState('Your JSON Output Will Display Here: ');
 
-  function updateShouldRunSUSHI(shouldRunSUSHI) {
+  function buttonFunctionality(shouldRunSUSHI, sushiOutput) {
```
cmoesel, line 29: I'd suggest using a more descriptive name... This is the Run button, right?  Maybe `handleRunButton` or `onRunButtonClick`?

src/App.js
```diff
@@ -23,29 +23,31 @@ export default function App() {
...
-            updateTextValue={updateTextValue}
-            updateShouldRunSUSHI={updateShouldRunSUSHI}
+            value={inputText}
+            updateTextValue={updateInputTextValue}
+            updateShouldRunSUSHI={buttonFunctionality}
           />
         </Grid>
         <Grid className={classes.itemTop} item xs={6}>
-          <JSONOutput shouldDisplaySUSHI={shouldRunSUSHI} text={text} />
+          <JSONOutput shouldDisplaySUSHI={shouldRunSUSHI} text={outputText} />
```
cmoesel, line 50: This is a minor comment, and probably a matter of preference, so you can choose to ignore it, but... is "should" really needed in the variable names?  Isn't this working kind of like a command?  Would `displaySUSHI` (and `runSUSHI`) be just as good (but shorter)?

src/components/JSONOutput.js
```diff
@@ -8,7 +8,9 @@ const useStyles = makeStyles((theme) => ({
     color: theme.palette.text.primary,
     background: theme.palette.grey[400],
     height: '100%',
-    fontFamily: 'Consolas'
+    fontFamily: 'Consolas',
+    gutterButtom: true,
```
cmoesel, line 12: Is this a typo?  Should it be `gutterBottom`?

src/App.js
SamCitron: A good point. I changed the shouldDisplaySUSHI to just displaySUSHI, however I think we should keep shouldRunSUSHI with the should in front of it as the name of our function to execute sushi is now runSUSHI(). We may just want to avoid confusion between the boolean value it represents and the actual calling of the function runSUSHI()?

cmoesel: That makes sense... That said, I feel like "should" implies that there is a chance it _won't_ though (like, "it _should_ run SUSHI, but really, who knows?!").  Maybe `doRunSushi` would be better?  I don't know -- I'm probably just getting nit-picky with semantics at this point!

src/components/JSONOutput.js
jafeltra: I guess a follow up question would be, if there was a typo, do we need that property at all?

PR #11, opened 2020-07-22
Fancier JSON display and doesn't remove output text on the right

SamCitron:
Fixes issue CIMPL-410 linked [here](https://standardhealthrecord.atlassian.net/secure/RapidBoard.jspa?rapidView=9&projectKey=INT&modal=detail&selectedIssue=CIMPL-410&assignee=5ece9e7747d31e0c2a0e7681). Makes a nicer, interactive output on the right hand side where the user can navigate the JSON output. Also ensures the previous JSON output won't disappear until the run button is clicked again and the refreshed output takes it's place.

### DIFF
--- a/src/components/RunButton.js
+++ b/src/components/RunButton.js
@@ -32,6 +32,7 @@ export default function RunButton(props) {
 
   //Sets the doRunSUSHI to true
   async function handleClick() {
+    props.onClick(true, 'Loading...', false);
     let isObject = true;
     const outPackage = await runSUSHI(props.text);
     let jsonOutput = JSON.stringify(outPackage, replacer, 2);
@@ -51,8 +52,6 @@ export default function RunButton(props) {
       jsonOutput = 'Your FSH is invalid. Just keep swimming!';
     }
 
-    //This double call to onClick just resets the doRunSUSHI variable so that react doesn't get confused with different incoming text
-    props.onClick(false, jsonOutput, isObject);
     props.onClick(true, jsonOutput, isObject);
   }
 
--- a/src/components/RunButton.js
+++ b/src/components/RunButton.js
@@ -32,6 +32,7 @@ export default function RunButton(props) {
 
   //Sets the doRunSUSHI to true
   async function handleClick() {
+    let isObject = true;
     const outPackage = await runSUSHI(props.text);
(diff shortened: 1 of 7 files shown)


cmoesel:
Cool! This is a big improvement!

I think when I looked at the React widget before, it supported showing the number of items in an array (without needing to expand it).  Is that something that's possible here?  If so, I think it would be helpful as it allows you to see at a glance how many profiles, extensions, etc there are...

src/components/RunButton.js
```diff
@@ -42,13 +43,17 @@ export default function RunButton(props) {
...
       }
     } else {
+      isObject = false;
       jsonOutput = 'Your FSH is invalid. Just keep swimming!';
     }
 
-    props.onClick(true, jsonOutput);
+    //This double call to onClick just resets the doRunSUSHI variable so that react doesn't get confused with different incoming text
+    props.onClick(false, jsonOutput, isObject);
+    props.onClick(true, jsonOutput, isObject);
```
jafeltra, line 56: I know we talked briefly about the need to have this call to `props.onClick` twice, but I was looking into why it was needed, and I can't figure out when it is needed. Do you have a specific series of steps that I can do in the app to reproduce the issue if the first call with `false` isn't there?

SamCitron: Sure! If you open up the page
1) Put in good FSH:
```
Profile:        FishPatient
Parent:         Patient
Id:             fish-patient
Title:          "Fish Patient"
Description:    "A patient that is a type of fish."
``` 
2) Don't reload, just backspace out some of that good FSH to make it bad FSH:
```
Profile:        FishPatient
Parent:         Patient
Id:             fish-patient
Title:          "Fish Patient"
Description:    "A patient that is
```
For me I get a React can't find error. Something about the parsing of the JSON will get messed up unless you reset it.

jafeltra: So I didn't exactly figure out why the error comes up, but I did have another idea. What if instead of the double call at the end of the function, we move it to the top. Then we could reset the text to say "Loading" which would give some indication that SUSHI is running! Something like this at the top of the function:

```
props.onClick(true, 'Loading...', false);
```

PR #12, opened 2020-07-24
Make console

SamCitron:
Fixes CIMPLE-408 linked [here](https://standardhealthrecord.atlassian.net/secure/RapidBoard.jspa?rapidView=9&projectKey=INT&modal=detail&selectedIssue=CIMPL-408&assignee=5ece9e7747d31e0c2a0e7681). Displays SUSHI logger in console on screen. However, displays all text at once, and is not yet color coded. Console clears with every run.

### DIFF
--- a/src/components/ConsoleComponent.js
+++ b/src/components/ConsoleComponent.js
@@ -9,6 +9,9 @@ const useStyles = makeStyles((theme) => ({
     background: theme.palette.common.black,
     height: '200%',
     fontFamily: 'Consolas'
+  },
+  pre: {
+    margin: '0px'
   }
 }));
 
@@ -25,7 +28,11 @@ export default function Console() {
     <Box className={classes.box} overflow="scroll">
       <h3>Console</h3>
       {msgArray.map((msg, i) => {
-        return <p key={i}>{msg}</p>;
+        return (
+          <pre key={i} className={classes.pre}>
+            {msg}
+          </pre>
+        );
       })}
     </Box>
   );
--- a/src/components/ConsoleComponent.js
+++ b/src/components/ConsoleComponent.js
@@ -7,16 +7,26 @@ const useStyles = makeStyles((theme) => ({
     padding: theme.spacing(2),
     color: theme.palette.common.white,
     background: theme.palette.common.black,
-    height: '100%',
+    height: '200%',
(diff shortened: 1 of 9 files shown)


src/components/ConsoleComponent.js
```diff
@@ -7,16 +7,27 @@ const useStyles = makeStyles((theme) => ({
     padding: theme.spacing(2),
     color: theme.palette.common.white,
     background: theme.palette.common.black,
-    height: '100%',
+    height: '200%',
```
jafeltra, line 10: Why did you need to change this height? Just curious!

src/components/TopBar.js
```diff
@@ -5,17 +5,20 @@ import { AppBar, Toolbar, Button, Typography } from '@material-ui/core';
...
-    margin: theme.spacing(2)
+    margin: theme.spacing(1),
+    color: 'white',
+    background: 'linear-gradient(45deg, #30638E 30%, #285a85 90%)',
+    boxShadow: '0 3px 5px 2px #285a85'
   },
   title: {
     flexGrow: 1,
-    edge: 'start',
-    fontFamily: 'Consolas'
```
jafeltra, line 18: Did you remove the `fontFamily` here as part of matching the top bar with FSH School? I'm just curious, especially since we were talking about the different fonts earlier.

src/components/ConsoleComponent.js
SamCitron: I just felt it looked better, without the increased height the console can be a bit narrow and you can't see entirely the box. However I can totally change it back if we would prefer it fit on one page

src/components/TopBar.js
SamCitron: Yes, I was trying to get it to most closely resemble the font used in the fsh-school or fsh-online.

jafeltra: Ah cool! Makes sense to me!

src/components/ConsoleComponent.js
jafeltra: I think maybe ideally everything would fit on one page without needing to scroll, but it needed to scroll before anyway, and I think you're right, this is better for now at least because you can see more of the console!

PR #20, opened 2020-09-03
FSH Syntax Highlighting

jafeltra:
This uses CodeMirror's [simple mode](https://codemirror.net/demo/simplemode.html) for defining a new language syntax highlighting. This mode uses regular expressions, similar to how our current VS Code extension works. Because of that, most of the regular expressions match the regular expressions in the VS Code extension. The tokens are used to categorize each regular expression and give them a specific color. I tried to match the colors based on the default color theme in VS Code. Pictures of a comparison are below.

The one difference between the regular expressions here and in the VS Code extension is the lack of look ahead and look behind. The simple mode approach for defining a language in Code Mirror can't support these. The purpose of the look behind/ahead is so that keywords that are not surrounded by spaces do not get highlighted. Without them, something like `code.coding.system` will highlight "system" in FSH Online but not in the VS Code extension. When implementing this, I decided it was easier to continue using the simple mode approach since it so closely mirrored the VS Code extension and this was only one small difference. If reviewers don't like the difference or if it becomes bothersome to users in the future, we can explore other ways to support the look behind.

VS Code:
image

FSH Online:
image

### DIFF
--- a/src/components/CodeMirrorComponent.js
+++ b/src/components/CodeMirrorComponent.js
@@ -1,12 +1,61 @@
 import React from 'react';
 import { Box } from '@material-ui/core';
-import { UnControlled as CodeMirror } from 'react-codemirror2';
+import { UnControlled as ReactCodeMirror } from 'react-codemirror2';
 import { makeStyles } from '@material-ui/core/styles';
+import CodeMirror from 'codemirror';
 import '../style/CodeMirrorComponent.css';
 import 'codemirror/lib/codemirror.css';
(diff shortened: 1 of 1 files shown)


cmoesel (approved):
Wow!  Looks awesome!

image

ngfreiter (approved):
Yeah this looks good. I think we do want to make this work better with stuff like "system", but I don't think it is worth holding up all syntax highlighting just because it is a little wonky in very specific cases.

PR #21, opened 2020-09-08
SUSHI Output updates

jafeltra:
This PR includes small changes requested to FSH Online to get it to alpha capabilities. These changes include:

- Including the list of errors at the top of the right hand side pane. Just the error messages are displayed but the full output can still be found in the console at the bottom of the page.
- Removing the snapshot from the StructureDefinitions in the package we get back from SUSHI
- Opening all fields in the JSON output by default. Now that the snapshot is not on the profiles and extensions, opening all the fields by default seems much more manageable and seems to be preferable.

Update: e66703c also updates the fonts used on the site. I just tried to align the style a bit more with FSH School, but the fonts are still slightly off, but I thought they were at least closer. If anyone is better with fonts than I am and is more bothered by the changes in this commit than master, I can revert it.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -24,7 +24,7 @@ let consoleMessages = [];
 let errorMessages = [];
 console.log = function getMessages(message) {
   consoleMessages.push(message);
-  if (message.startsWith('error')) {
+  if (message && message.startsWith('error')) {
     errorMessages.push(message);
   }
 };
@@ -47,6 +47,7 @@ export default function App() {
     setOutputText(sushiOutput);
     setIsOutputObject(isObject);
   }
+
   function updateInputTextValue(text) {
     setInputText(text);
   }
--- a/src/App.js
+++ b/src/App.js
@@ -21,11 +21,11 @@ const useStyles = makeStyles((theme) => ({
 
 const log = console.log; //eslint-disable-line no-unused-vars
 let consoleMessages = [];
-let errorMessages = [];
+let errorAndWarningMessages = [];
 console.log = function getMessages(message) {
   consoleMessages.push(message);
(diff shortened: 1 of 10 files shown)


src/App.js
```diff
@@ -20,9 +20,13 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 const log = console.log; //eslint-disable-line no-unused-vars
-let msgArray = [];
+let consoleMessages = [];
+let errorMessages = [];
 console.log = function getMessages(message) {
-  msgArray.push(message);
+  consoleMessages.push(message);
+  if (message && message.startsWith('error')) {
```
ngfreiter, line 27: I think it might be worthwhile to also show warnings, since that aligns well with the summary given by the box at the end of SUSHI.

ngfreiter (commented):
The changes work well for me!

src/App.js
jafeltra: Okay! I was thinking about including them, but couldn't decide if it was too much information to have immediately available or if it would be helpful. But I'd say an additional vote for helpful tips the scales!

jafeltra:
Update again: 08afde8 includes a fix for using the monospace font for the `pre` tag, which all the console output is inside of. This made it so the summary box (with the error total and pun) lined up better on Mac. If someone with a Windows computer could make sure the box still lines up well, that would be great. If it doesn't, I can revert the change.

src/App.js
jafeltra: Included in 7383c11!

ngfreiter (approved):
Box still looks good. Warnings look good. Looks good!

PR #22, opened 2020-09-09
Update SUSHI version

jafeltra:
This PR updates FSH Online to use the latest version of SUSHI. It also updates the import statements to use the same method as GoFSH and includes the latest puns from SUSHI.

### DIFF
--- a/src/tests/utils/Load.test.js
+++ b/src/tests/utils/Load.test.js
@@ -1,8 +1,10 @@
 import { unzipDependencies, loadDependenciesInStorage, loadAsFHIRDefs } from '../../utils/Load';
-import { FHIRDefinitions } from 'fsh-sushi/dist/fhirdefs/FHIRDefinitions';
+import { fhirdefs } from 'fsh-sushi';
 import tarStream from 'tar-stream';
 import http from 'http';
 import 'fake-indexeddb/auto';
+
+const FHIRDefinitions = fhirdefs.FHIRDefinitions;
 
 describe('#unzipDependencies', () => {
   let getSpy = jest.SpyInstance;
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -10,6 +10,10 @@ module.exports = function override(config, env) {
       use: 'null-loader'
     },
     {
+      test: /.*readline-sync.*/,
+      use: 'null-loader'
+    },
+    {
       test: /.*tar(\\|\/).*/,
       use: 'null-loader'
     },
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,9 +13,9 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.10.5",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.10.5.tgz",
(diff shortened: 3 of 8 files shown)


src/tests/utils/Load.test.js
```diff
@@ -1,9 +1,11 @@
 import { unzipDependencies, loadDependenciesInStorage, loadAsFHIRDefs } from '../../utils/Load';
-import { FHIRDefinitions } from 'fsh-sushi/dist/fhirdefs/FHIRDefinitions';
+import { fhirdefs } from 'fsh-sushi';
 import tarStream from 'tar-stream';
 import http from 'http';
 import 'fake-indexeddb/auto';
 
+const FHIRDefinitions = fhirdefs.FHIRDefinitions;
```
ngfreiter, line 7: Could you do this more simply by changing the import to:
```
import { FHIRDefinitions } from 'fsh-sushi/fhirdefs';
```
I think that should work the same way, and as far as I know that's the more conventional way to do this. And I think that is what we do in GoFSH as well.

ngfreiter: I notice this in some other places as well, so same question goes for those other places.

ngfreiter (commented):
One kinda annoying question from me.

src/tests/utils/Load.test.js
cmoesel: No, IIRC, you actually _can't_ do that.  Because `fsh-sushi/fhirdefs` is not an actual file in the package.  `fhirdefs` is an exported variable from the package's `index.js`.

cmoesel: That said, I'd love to be wrong about this.

ngfreiter: You are not wrong about this, @jafeltra, please ignore my kinda annoying question.

ngfreiter:
But don't worry I have more questions. When I do `npm start`, I get the following page:
image
 
Is this happening to anyone else? Or just a me problem?

cmoesel:
I've run into two issues.

1. NPM reports high criticality security vulnerabilities.  If you run `npm audit` it will tell you how to fix them (or you can just run `npm audit fix`).  It would be nice to go ahead and do that.

2. When I try to actually run this locally via `npm run`, it crashes for me:
image

(Oh -- I see that Nick just beat me to the punch on that last one).

cmoesel:
BTW -- I tried w/ Node 10 and Node 12 -- and I tried blowing away node_modules and re-installing.  No dice no matter what.

jafeltra:
Oh huh. Maybe I had the app still running because I hadn't noticed this, but I'm seeing it now. I'll fix those two things.

jafeltra:
The reason the app was crashing was because we added the `readline-sync` package in one of the newer versions of SUSHI than we previously used here. That is another package that cannot be used in the browser. Since it was only used during the `sushi --init` command, I've just added this package to list of other packages we choose to not load in the browser. This is addressed in 67e86fb.

I've also addressed the vulnerabilities in c8b08a7.

src/tests/utils/Load.test.js
jafeltra: Okay I shall ignore. I believe Chris is correct at to why we can't do it that way. I tried to stick with the way GoFSH does it, although GoFSH does not reassign the imports to variables, it just uses things like `fhirdefs.FHIRDefinitions` directly, so I could swap what I did here to mirror that more closely, if preferred.

ngfreiter (approved):
Works for me now!

PR #24, opened 2020-09-23
User Provided Configuration

jafeltra:
This PR supports basic user controls for SUSHI configuration. I believe the only SUSHI config options that should be applicable to the playground are `canonical`, `version`, and `dependencies`. This PR allows a user to specify the `canonical` and `version` and the field will be used by SUSHI when the "Run" button is clicked. Since we don't yet support loading any other dependencies besides R4, I added a disabled field for those, although it can be removed or commented out if that is easier. Also, if there is any feedback on the format/location of these controls, they can always be changed. The dialog box was my initial idea and it looked alright, but I'm no designer.

There is a warning in the console that gets logged by the Material UI Dialog component. I believe it is just a warning due to the fact that we are using React Strict Mode, and I found a [GitHub Issue](https://github.com/mui-org/material-ui/issues/13394) where Material UI seems to be tracking support for Strict Mode in the next major version, so I think it is okay to leave this as is. The warning does not get logged in the production build that will end up on fshschool.org.

Also, all I did was rename the RunButton.js -> SUSHIControls.js. I'm not sure why GitHub seems so thrown off by the rename + additions. f0cb64f has all the functional changes and 3a5d73d does the renaming if that's easier to look at.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -5,7 +5,7 @@ import TopBar from './components/TopBar';
 import JSONOutput from './components/JSONOutput';
 import ConsoleComponent from './components/ConsoleComponent';
 import CodeMirrorComponent from './components/CodeMirrorComponent';
-import RunButton from './components/RunButton';
+import SUSHIControls from './components/SUSHIControls';
 
 const useStyles = makeStyles((theme) => ({
   container: {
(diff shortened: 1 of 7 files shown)


ngfreiter (commented):
This works well for me, but I think one improvement would be if the `canonical` value you've set persists. So if I open the configuration, change the `canonical` to `nick.org`, run, and then realize I actually want the `canonical` to be `nicks-canonical.org`, when I re-open the configuration, I'd expect the `canonical` field to currently still say `nick.org`, not to revert back to the default.

jafeltra:
That's a great idea that I didn't even think about. f42760a uses the state variable as the default, so your values should persist if you close and reopen the modal. They won't persist over refreshes, but neither does the FSH text, so I figured that was okay for now.

ngfreiter (approved):
This is looking good to me!

cmoesel (approved):
Super sweet.  Nicely done!

PR #25, opened 2020-09-23
Documentation and Actions updates

jafeltra:
Updates the README with basic description of the app and basic FHIR Shorthand information. I also simplified the actions on the repo.


ngfreiter (commented):
This looks good to me, but I'm just wondering why the simplification of the Github actions?

jafeltra:
Well my main motivation was that I believe @cmoesel mentioned once that we don't need them all, and I was thinking that we didn't need to test on all the various OS and Node versions since the code will be built down to plain JS. But now I'm realizing that the test and lint checks don't operate on the built code, so maybe that was an incorrect thought. I can do some research (or talk to Chris again) and potentially revert that commit.

jafeltra:
Ah okay I've consulted @cmoesel and he has enlightened me. We don't need all the various options because we are the ones deploying the site, so we know the environment we are using. Since the deploy workflow is using Node 12 and ubuntu-latest, we really only need to run the actions on that environment.

ngfreiter (approved):
Sounds good!

cmoesel (approved):
Looks good.  I updated the branch settings to allow PRs to be merged as long as they pass these two jobs.  (Prior to that, this PR was not happy because it did not run jobs that were required for merging).

PR #29, opened 2021-01-05
Permalink

SamCitron:
Allows users to share their FSH with others by generating a short url. Gives users the option to copy the link to their clipboard, and the app will render properly with the shared FSH. Addresses issue: https://standardhealthrecord.atlassian.net/secure/RapidBoard.jspa?rapidView=9&modal=detail&selectedIssue=CIMPL-610&assignee=5ece9e7747d31e0c2a0e7681

Note: a few of the tests are unfinished. Julia and I will likely return to them over the week or next week when I am back.


cmoesel:
This is such a cool feature!  I haven't closely reviewed the code, but I have a few thoughts from a user perspective:

1. Stylistically, the actual shortened link is kind of difficult to read.  Can we darken it?
image

2. The actual link is `https://bit.ly/35dodLE`, which works, but also loses all context and branding (it's not obviously a FSHOnline link).  Could we actually make it work so that when the bitly API gives us a link like above, we present it to the user as `https://fshschool.org/FSHOnline/35dodLE`?  Then when someone navigates to that link, we can pull off the `35dodLE` part and use the bitly `expand` function to get the full URL, then decode the FSH from the path part of that URL?  Then our permalinks would truly look like _our_ permalinks (and they would work w/ localhost too).

And a few from an engineering perspective:

3. Are we sure that using the URL format `https://fshschool.org/FSHOnline/{Base64orShortCode}` will actually work when we deploy it?  It seems like a fancy router feature, and it works when you use `npm start`, but I don't see how GitHub pages will know _not_ to try to serve a physical page from that location.  To test this, I used `npm run build` to build the static site (like will be used w/ GH pages), and then deployed it under `/FSHOnline` using a local http server, and when I tried to go to a link w/ the base64 part, I got a 404:
> [2021-01-06T14:20:22.462Z]  "GET /FSHOnline/UHJvZmlsZTogTXlDb29sUGF0aWVudApQYXJlbnQ6IFBhdGllbnQKKiBhY3RpdmUgPSB0cnVlCiogbmFtZSAxLi4xIE1TCg==" Error (404): "Not found"

4. If we can get that format of URL to work, is putting the shortlink/base64 directly after `https://fshschool.org/FSHOnline/` future-proof?  If we want to add other features/pages (for example, `https://fshschool.org/FSHOnline/about`, is that possible or does it cause problems w/ routing?  Might it be better to do something like `https://fshschool.org/FSHOnline/share/35dodLE`? (Or if that style of link doesn't work at all, maybe `https://fshschool.org/FSHOnline/?share=35dodLE`?)

5. I noticed you are not using any compression before you base64 encode the FSH.  Have you tested the limits to see how much FSH you can share before things break?  Is compression not necessary?

jafeltra:
After talking to Sam at the beginning of the week and then stepping away for a few days, I took another look at the two tests that we had gotten stuck on and I think I came up with work arounds for them.

- a90dd72 changes up the approach we took for testing if the copy to clipboard function was working as expected. Instead of trying to look at the setState method, I decided to look for the button text that would be rendered if the state was updated as we expected. I think this is similar to other approaches we have taken in the past. I also mocked out the copy-to-clipboard module since we don't need to actually be calling or testing that and it was causing errors.
- a5a835e restructures the routes a bit and added a test that the basic app loaded. It also mocks out the actual App component and instead just tests that the appropriate mock is rendered when we visit a path. I think we had tried various combinations of restructuring the routes earlier in the week, but maybe it didn't work without the mocks.

jafeltra:
Okay one last change from me here! 6ec06af uses an environment variable to get the API Key for Bitly requests. The repository should be set up to use this variable when this branch is merged and the code is deployed!

src/App.js
```diff
@@ -29,14 +31,35 @@ console.log = function getMessages(message) {
...
+  let decodedText = null;
   const [doRunSUSHI, setDoRunSUSHI] = useState(false);
-  const [inputText, setInputText] = useState('Edit FSH here!');
+  const [inputText, setInputText] = useState('Edit FSH Here!');
   const [outputText, setOutputText] = useState('Your JSON Output Will Display Here: ');
   const [isOutputObject, setIsOutputObject] = useState(false);
 
+  useEffect(() => {
+    async function waitForFSH() {
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
jafeltra, line 56: Do you know what we need to disable this lint rule? Do you know if there's a way to change how we write this function to avoid disabling the lint rule?

I did look at this a bit and it looks possible, but maybe I'm missing an edge case, so we can talk about this one.

src/components/SUSHIControls.js
```diff
@@ -70,17 +83,48 @@ export function sliceDependency(dependencies) {
...
+  };
+
+  const handleOpenShare = async () => {
+    const encoded = deflateSync(props.text).toString('base64');
+    const longLink = `https://fshschool.org/FSHOnline/share/${encoded}`;
+    const bitlyLink = await generateLink(longLink);
+    if (bitlyLink.errorNeeded === true) {
+      handleOpenShareError();
+    } else {
+      const bitlySlice = bitlyLink.link.slice(15);
```
jafeltra, line 110: Can you add a comment to this line reminding future-us what the 15 represent in this slice?

jafeltra (commented):
This is looking really good to me! I think all of Chris's comments are addressed, but I will let him confirm as well. I added a few questions I had below!

cmoesel (changes requested):
I'm not a React guru, but I didn't see anything blatantly wrong in the code.

But... I did notice a problem that is on this branch, but not on the live FSHOnline -- so I *think* it must have been introduced here.  Basically, if you type some FSH and then try to edit text in the middle of the FSH content, it will let you type one character and then it will move the cursor to the end!  So, I think we need to fix that before we can merge.

Also, not your fault at all, but when I tried sharing w/ real content (from mCODE), I got way less than 500 lines.  It couldn't even share 100 lines.  So I guess it depends on the FSH.  I think maybe before it was working well for me because I was copy/pasting the same text over and over.  Maybe zlib has a special algorithm that handles that well.  Anyway, I think it's still the best we can do, but... we may struggle with the limit.

cmoesel (dismissed):
This looks great and the hash router works!  How exciting!

My only wish is that we knew what the URL limit was for bitly, but alas, I couldn't find it in the docs either.  And even if we did, it would be hard to translate it into something meaningful for the user (since it really depends on how well their particular FSH compresses).  I guess we'll just take a wait-and-see approach on this and evaluate if we need to make changes later.

src/App.js
```diff
@@ -29,14 +31,35 @@ console.log = function getMessages(message) {
...
-  const classes = useStyles();
+export async function decodeFSH(encodedFSH) {
+  if (encodedFSH.text === undefined) {
+    return 'Edit FSH here!';
+  } else {
+    const promisedURL = await expandLink(encodedFSH);
+
+    // Removes the encoded data from the end of the url, starting at index 38
+    const sliced64 = promisedURL.long_url.slice(40);
+    const displayText = inflateSync(Buffer.from(sliced64, 'base64')).toString('utf-8');
```
jafeltra, line 42: I think I just hit a random edge case, but I had some nonsense typed into the id part of `.../share/:id` and the `promisedURL.long_url` I got back was not a FSH Online URL, so then I got an error when we try to slice and decode the text (error below). I think we might want to guard against this case, even if it is rare (you'd only get it if you were typing in a url and made a mistake probably). Maybe we can check that the `promisedURL.long_url` starts with `https://fshschool.org/FSHOnline/#/share/` or that `sliced64` isn't an empty string. Or maybe both. In those cases, we probably just want to return the default "Edit FSH here!" string. Either that, or maybe some of this manipulation should happen in the `expandLink` function in the `then`.

```
index.js:352 Uncaught (in promise) Error: unexpected end of file
    at Zlib._handle.onerror (index.js:352)
    at Zlib.push../node_modules/browserify-zlib/lib/binding.js.Zlib._error (binding.js:283)
    at Zlib.push../node_modules/browserify-zlib/lib/binding.js.Zlib._checkError (binding.js:238)
    at Zlib.push../node_modules/browserify-zlib/lib/binding.js.Zlib._write (binding.js:116)
    at Zlib.push../node_modules/browserify-zlib/lib/binding.js.Zlib.writeSync (binding.js:77)
    at Inflate.push../node_modules/browserify-zlib/lib/index.js.Zlib._processChunk (index.js:511)
    at zlibBufferSync (index.js:236)
    at push../node_modules/browserify-zlib/lib/index.js.exports.inflateSync (index.js:161)
    at decodeFSH (App.js:44)
    at async waitForFSH (App.js:60)
```

jafeltra (commented):
This looks really good. I tried out a few things in the browser and found one issue we probably should address. Besides that, it looks great! Let me know if you want to chat about this at all.

jafeltra (approved):
This looks really great! This is such a cool feature. I agree that it would be nice to know the length of FSH we could share, but hopefully we can just evaluate that need as it arises.

cmoesel (approved):
Still looks good.  Thanks!

PR #30, opened 2021-01-12
Object Store Clean Up

SamCitron:
Deletes empty ObjectStores that users have created in the past, the old 'resources' object store, and deletes any empty ObjectStores that users may create by entering unrecognized dependencies. Addresses comments made here: https://github.com/FSHSchool/FSHOnline/pull/28#issuecomment-754134629.

Note: there is still one test in `Load.test.js` that must be fixed or determined unnecessary.

### DIFF
--- a/src/tests/utils/Load.test.js
+++ b/src/tests/utils/Load.test.js
@@ -32,7 +32,7 @@ describe('#unzipDependencies', () => {
     const callbackFunction = getSpy.mock.calls[0][1];
     callbackFunction({ statusCode: 404 });
     expect(getSpy).toBeCalled();
-    expect(unzipPromise).resolves.toEqual({ emptyDependencies: ['hello123'], resourceArr: resources });
+    await expect(unzipPromise).resolves.toEqual({ emptyDependencies: ['hello123'], resourceArr: resources });
   });
 });
 
--- a/src/tests/utils/Load.test.js
+++ b/src/tests/utils/Load.test.js
@@ -8,6 +8,7 @@ const FHIRDefinitions = fhirdefs.FHIRDefinitions;
 
 describe('#unzipDependencies', () => {
   let getSpy = jest.SpyInstance;
+  // let getSpyBad = jest.SpyInstance;
   let tarSpy = jest.SpyInstance;
   let resources = [];
   beforeAll(() => {
@@ -29,6 +30,20 @@ describe('#unzipDependencies', () => {
     expect(getSpy).toBeCalled();
     expect(getSpy).toBeCalledWith('https://packages.fhir.org/hl7.fhir.r4.core/4.0.1', callbackFunction);
   });
+
+  //TODO - test failed http request
+
+  // it('should add failed http requests to a list of emptyDependencies', async () => {
+  //   getSpyBad = jest.spyOn(http, 'get').mockImplementation(() => {
+  //     const res = { statusCode: 404 };
+  //     return res;
+  //   });
+  //   const unzipReturn = await unzipDependencies(resources, 'hello', '123');
+  //   const callbackFunction = getSpy.mock.calls[0][1];
+  //   expect(getSpyBad).toBeCalled();
+  //   expect(getSpyBad).toBeCalledWith('https://packages.fhir.org/hello/123', callbackFunction);
+  //   expect(unzipReturn).toBe({ resourceArr: resources, emptyDependencies: ['hello#123'] });
+  // });
 });
 
 describe('#loadDependenciesInStorage', () => {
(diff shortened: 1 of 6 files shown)


src/tests/utils/Load.test.js
```diff
@@ -29,6 +30,20 @@ describe('#unzipDependencies', () => {
...
+  //   getSpyBad = jest.spyOn(http, 'get').mockImplementation(() => {
+  //     const res = { statusCode: 404 };
+  //     return res;
+  //   });
+  //   const unzipReturn = await unzipDependencies(resources, 'hello', '123');
+  //   const callbackFunction = getSpy.mock.calls[0][1];
+  //   expect(getSpyBad).toBeCalled();
+  //   expect(getSpyBad).toBeCalledWith('https://packages.fhir.org/hello/123', callbackFunction);
+  //   expect(unzipReturn).toBe({ resourceArr: resources, emptyDependencies: ['hello#123'] });
+  // });
```
ngfreiter, line 46: I think I've figured out a valid way to test this. If someone else knows a better way, definitely still open to hearing it, but this is what I've got:
```javascript
it('should add failed http requests to a list of emptyDependencies', () => {
    const unzipPromise = unzipDependencies(resources, 'hello', '123');
    // Get the callback, and run it with a 404
    const callbackFunction = getSpy.mock.calls[0][1];
    callbackFunction({ statusCode: 404 });
    expect(getSpy).toBeCalled();
    expect(unzipPromise).resolves.toEqual({ emptyDependencies: ['hello123'], resourceArr: [] });
  });
```
Note that this code will require you remove the `tarSpy`, but as far as I could tell the `tarSpy` was not necessary anyway. What this code is doing is mocking the entire `http.get`, and then since we can get the arguments that a function is called with, we get the second argument, which is actually a callback function. Then we manually trigger that callback with a 404 status code, and we can test that the `emptyDependencies` array is indeed populated.

src/tests/utils/Load.test.js
```diff
@@ -1,34 +1,39 @@
...
     expect(getSpy).toBeCalled();
     expect(getSpy).toBeCalledWith('https://packages.fhir.org/hl7.fhir.r4.core/4.0.1', callbackFunction);
   });
+
+  it('should add failed http requests to a list of emptyDependencies', async () => {
+    const unzipPromise = unzipDependencies(resources, 'hello', '123');
+    const callbackFunction = getSpy.mock.calls[0][1];
+    callbackFunction({ statusCode: 404 });
+    expect(getSpy).toBeCalled();
+    expect(unzipPromise).resolves.toEqual({ emptyDependencies: ['hello123'], resourceArr: resources });
```
jafeltra, line 35: I think this last expect statement might need to be `await`ed. I'm not 100% I understand why, except that I tried commenting out the line that pushes to the `emptyDependencies` array and the test still passed, although it did show that the two objects didn't match. Awaiting the last expect does have the test pass and fail as expected.

src/tests/utils/RunSUSHI.test.js
```diff
@@ -50,6 +74,7 @@ describe('#runSUSHI', () => {
     const input = 'Improper FSH code!';
     const outPackage = await runSUSHI(input, defaultConfig);
     expect(loadSpy).toHaveBeenCalled();
+    expect(checkForDatabaseUpgradeSpy).toHaveBeenCalled();
```
jafeltra, line 77: Hey cool! Bonus tests!

src/utils/Processing.js
```diff
@@ -22,6 +63,9 @@ export function checkForDatabaseUpgrade(dependencyArr, databaseName = 'FSH Playg
       database = event.target.result;
       let existingObjectStores = database.objectStoreNames;
       helperReturn.version = database.version;
+      if (existingObjectStores.contains('resources')) {
+        helperReturn.shouldUpdate = true;
```
jafeltra, line 67: The answer to this might be yes, but is there a need to keep checking the other if/else case if we hit this case? Or can we make this part of the same if case as the other two and have this close the database and resolve if we reach this first case?

jafeltra (commented):
This looks really good! It works really nicely! I left a few comments inline below. I also think I might have found two edge cases we might want to address:

- When I have an empty object store already in my IndexedDB that got created before I switched to this branch, it doesn't get deleted. It looks like there might be code trying to cover this case in `cleanDatabase`. Is that something you were trying to cover?
- When I had a "resources" object store left over in my database and then tried to run SUSHI on this branch with dependencies specified, I get an error. My disclaimer for this one is that I was testing out a bunch of different cases when I get this error so I'm not 100% sure what caused it, so I'll try to narrow it down, but I do think we'll want to fix this one.

jafeltra (approved):
This looks good to me! I tested out with a database that had `resources`, a filled dependency and an empty dependency store. Then I switched to this branch and tested running with and without dependencies listed in the config menu. Both cases correctly downloaded new dependencies, removed resources, removed old empty dependencies, and removed current empty dependencies.

I'd say this looks good! We'll still want a second reviewer though.

ngfreiter (approved):
This looked good to me, thanks!

PR #31, opened 2021-01-22
Examples

SamCitron:
Creates examples button, page, and functionality. Populates examples page dynamically from examples-config.json. 
Addresses this [issue](https://standardhealthrecord.atlassian.net/secure/RapidBoard.jspa?rapidView=9&modal=detail&selectedIssue=CIMPL-609&assignee=5ece9e7747d31e0c2a0e7681) on Jira.

We will need to determine which examples to provide and group titles before merging and deploying.


src/App.js
```diff
@@ -50,7 +51,8 @@ export async function decodeFSH(encodedFSH) {
 
 export default function App(props) {
   const classes = useStyles();
-  const text64 = props.match.params;
+  const urlParam = props.match.params;
+  const url = window.location.href;
```
jafeltra, line 55: Do you know if there's a way to use `props.match` here similar to how we do with the Share functionality? I'm not actually sure if there's a difference, but it seems nice for consistency, and we wouldn't have to use `window`.

src/components/SUSHIControls.js
```diff
@@ -176,6 +199,44 @@ export default function SUSHIControls(props) {
...
+    let columns = [];
+    let components = [];
+    let obj = Object.entries(config);
+    for (let group = 0; group < obj.length; group++) {
+      components.push([]);
+      for (let file = 0; file < obj[group][1].files.length; file++) {
+        let fileName = obj[group][1].files[file].name;
+        let fileLink = obj[group][1].files[file].link;
+        components[group].push(
+          <Grid item xs>
```
jafeltra, line 212: I'm getting a warning in the console that says `Warning: Each child in a list should have a unique "key" prop.`. I think that is because these `Grid` items are coming from an array. There's a bit of documentation about it [here](https://reactjs.org/docs/lists-and-keys.html#keys). I think we can add a "key" prop and it should be happier.

src/components/SUSHIControls.js
```diff
@@ -176,6 +199,44 @@ export default function SUSHIControls(props) {
...
+              {fileName}
+            </Link>
+          </Grid>
+        );
+      }
+    }
+    for (let col = 0; col < obj.length; col++) {
+      let groupName = obj[col][0];
+      columns.push(
+        <Grid item container direction="column" alignItems="flex-start" xs={3} spacing={3}>
```
jafeltra, line 229: Similar to above, I think we need a "key" to resolve the warning.

src/components/SUSHIControls.js
```diff
@@ -264,6 +328,27 @@ export default function SUSHIControls(props) {
...
+          aria-labelledby="form-dialog-title"
+          maxWidth="lg"
+          fullWidth
+          classes={{ paper: classes.dialogPaper }}
+        >
+          <DialogTitle id="form-dialog-title">Examples</DialogTitle>
+          <DialogContent>
+            <DialogContentText>Use our pre-created examples to learn FSH and get swimming!</DialogContentText>
+            <Grid container direction="row" justify="space-between" alignItems="flex-start" spacing={0}>
+              <React.Fragment>{ExampleGrid()}</React.Fragment>
```
jafeltra, line 343: I've never seen this `React.Fragment` before. I'll look up some documentation about it! There is also potentially a way you can do this without calling a function for the grid, but we can talk about that and decide if it's something we want to do now.

jafeltra (commented):
This looks really awesome! I had a few small things below that we can talk about!

src/tests/components/SUSHIControls.test.js
```diff
@@ -289,3 +290,31 @@ it('shows an error when the FSH file is too long to share', async () => {
...
+  act(() => {
+    const exampleBtn = getByText('Examples');
+    fireEvent.click(exampleBtn);
+  });
+  const textElement = getByText(/Use our pre-created examples to learn FSH and get swimming!/i);
+  const textElement1 = getByText(/Group 1/i);
+  const textElement2 = getByText(/Group 2/i);
+  const textElement3 = getByText(/Hello World/i);
+  const textElement4 = getByText(/Start Swimming/i);
+  const textElement5 = getByText(/Veternarian/i);
```
jafeltra, line 312: I was thinking about this again and this test will fail as soon as we fill in real examples. I'm thinking it might be worth it to pass in `config` as a prop to `SUSHIConfig` so that we can mock out the config in the test here. Then we'll have full control over the test and things won't break when we change the examples.

The other idea would be if we can mock out that import here, and then we don't have to change the component props at all.

We can also talk about if this is one we want to do now or later!

jafeltra (commented):
Oops, I thought of one more thing.

jafeltra (approved):
This looks really awesome! Thanks for making those updates!

This looks good to me, so I'm approving! If another reviewer notices anything next week, I can make the updates.

ngfreiter (commented):
Just had some small suggestions

src/components/SUSHIControls.js
```diff
@@ -264,6 +327,27 @@ export default function SUSHIControls(props) {
...
+          <DialogTitle id="form-dialog-title">Examples</DialogTitle>
+          <DialogContent>
+            <DialogContentText>Use our pre-created examples to learn FSH and get swimming!</DialogContentText>
+            <Grid container direction="row" justify="space-between" alignItems="flex-start" spacing={0}>
+              <ExampleGrid />
+            </Grid>
+          </DialogContent>
+          <DialogActions>
+            <Button onClick={handleCloseExamples} color="primary">
+              Done
```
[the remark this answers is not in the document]
jafeltra: Makes sense to me!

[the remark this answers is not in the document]
jafeltra: I lied. I chose "Close" because there wasn't really an action that the button was cancelling. Thoughts?

src/App.js
```diff
@@ -3,11 +3,13 @@ import { inflateSync } from 'browserify-zlib';
 import { makeStyles } from '@material-ui/core/styles';
 import { Grid } from '@material-ui/core';
 import { expandLink } from './utils/BitlyWorker';
+import { setExampleText } from './utils/ParseExamples';
 import TopBar from './components/TopBar';
 import JSONOutput from './components/JSONOutput';
 import ConsoleComponent from './components/ConsoleComponent';
 import CodeMirrorComponent from './components/CodeMirrorComponent';
 import SUSHIControls from './components/SUSHIControls';
+import config from './examples/examples-config.json';
```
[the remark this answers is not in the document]
jafeltra: I added some documentation to the README, but I didn't add anything to the file itself since it is a JSON file and doesn't support comments. If you think it would be helpful to have the documentation closer to the actual file, I could convert it to just a `.js` file with an object or I could add a `examples-config.example.json`. Let me know what you think of those options and with the content of what I added!

ngfreiter (approved):
Looks good to me!

src/App.js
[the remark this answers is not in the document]
ngfreiter: Nah, readme is fine, looks good!

PR #32, opened 2021-01-26
Update to SUSHI v1.1.0

jafeltra:
This PR:
- updates the version of SUSHI to v1.1.0
- uses SUSHI's newly available `getRandomPun` function to use a random pun
- updates the hardcoded SUSHI version visible in the app. There's a separate task to investigate doing this in a different way.


cmoesel (approved):
Splendid!

ngfreiter (approved):

PR #33, opened 2021-01-26
Bug fix for incorrect error count in console

jafeltra:
There was a bug I've noticed where the number of errors and warnings were incorrect in the Console section. There was some special logic for calculating the differences in number of errors between runs. I replaced this with just `stats.reset()`. I was originally going to use the `ErrorsAndWarnings` class that `fshToFHIR` uses, but I really only needed the number of errors and warnings. I know the `stats.reset` function has been around for a while, but I can't find any reason why we wouldn't want to use it, so I'm guessing we must have just missed it when implementing this originally.

In order to replicate this bug on fshschool.org/FSHOnline (or `master`), do the following:

- Enter some invalid FSH and run SUSHI
   ```
    Profile: A
    Parent: B
    ```
    There should be 1 error reported
- Fix the FSH so it is valid and run SUSHI. There should be zero errors reported.
- Change the FSH to make it invalid again and run SUSHI. There should be an incorrect number of errors in the console.

Repeating those steps on this branch should always have the correct number of errors, even at that third step. I didn't really bother trying to figure out what was going wrong with the math because I thought resetting was the right approach anyhow.

### DIFF
--- a/src/utils/RunSUSHI.js
+++ b/src/utils/RunSUSHI.js
@@ -12,14 +12,6 @@ const Type = utils.Type;
 const FHIRDefinitions = fhirdefs.FHIRDefinitions;
 
 /**
- * TODO: Keep a running tally of errors because the logger stats do not reset between SUSHI runs
- * If we have a way to clear the stats in the logger, these variables should be removed and we can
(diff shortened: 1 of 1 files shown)


cmoesel (approved):
Love it.

PR #35, opened 2021-02-09
Update to SUSHI v1.2.0

### LINKED ISSUE #34: Broken on Safari
Oof.  FSH Online doesn't work _at all_ in Safari.  Just a white screen.
image

And here's the console:
image

jafeltra:
This PR updates the installed SUSHI to the latest version, v1.2.0. This SUSHI update has a fix for using regex lookbehind assertions, and it fixes #34 so the app will load in Safari.

### DIFF
--- a/src/components/SUSHIControls.js
+++ b/src/components/SUSHIControls.js
@@ -215,7 +215,7 @@ export default function SUSHIControls(props) {
               margin="dense"
               fullWidth
               label="Dependencies"
-              helperText="Format: packageId#version, packageId#version (ex. hl7.fhir.us.core#3.1.1)"
+              helperText="Format: packageId#version, packageId#version (e.g., hl7.fhir.us.core#3.1.1)"
               defaultValue={dependencies}
               onChange={updateDependencyString}
             />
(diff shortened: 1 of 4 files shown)


jafeltra:
dfc3be6 also updates the helper text under the "Dependencies" field to include US Core as an example and change the wording to `packageId`, based on feedback from @markkramerus. This was such a small change I just decided to lump it in here.

src/components/SUSHIControls.js
```diff
@@ -215,7 +215,7 @@ export default function SUSHIControls(props) {
               margin="dense"
               fullWidth
               label="Dependencies"
-              helperText="dependencyID#version, dependencyID#version"
+              helperText="Format: packageId#version, packageId#version (ex. hl7.fhir.us.core#3.1.1)"
```
cmoesel, line 218: This is a nit, but isn't it customary to use `e.g.,` in a situation like this rather than `ex.`?

jafeltra: I almost used that, so I'm happy to swap it out!

jafeltra: 21b4522 swapped!

ngfreiter (approved):
This looks good, and I tested features from 1.2.0, and they seem to be working.

PR #36, opened 2021-03-02
Initial GoFSH Implementation

jafeltra:
This adds the initial functionality for running GoFSH in FSH Online. It sets up the function we will use to run GoFSH given FHIR JSON, which follows the `FhirToFsh` function very closely. There are a few hardcoded objects in the code right now that I've been using to test. The dependencies for the example IG included, as well as any dependencies in the Configuration modal are loaded and used.

This also includes the `npm pack`ed GoFSH  now that the changes to support optimizers are in but it is not yet released. Once we release a new version, I'll add it as a regular dependency and remove the .tgz.

I wanted to move this code into the more central `gofsh` branch, and also get any initial feedback on the approach as I continue to work on other features this sprint.


src/utils/RunSUSHI.js
```diff
@@ -11,6 +13,73 @@ const getRandomPun = utils.getRandomPun;
...
+ * dependencies: user set, defaults to []
+ * @param {FHIRDefinitions} testDefs - this should only be used by the unit tests so they can provide their own definitions.
+ * @returns {string} the FSH
+ */
+export async function runGoFSH(input, options, testDefs = null) {
+  // Read in the resources, either as JSON objects or as strings
+  const docs = []; // WildFHIR[]
+  input.forEach((resource, i) => {
+    const location = `Input_${i}`;
+    if (typeof resource === 'string') {
```
ngfreiter, line 46: Can the input be anything other than strings here?

src/utils/RunSUSHI.js
```diff
@@ -11,6 +13,73 @@ const getRandomPun = utils.getRandomPun;
...
+  // Set up the FHIRProcessor
+  const lake = new processor.LakeOfFHIR(docs);
+  let defs = testDefs || new FHIRDefinitions(); // The tests pass in their own set of definitions to use
+  const fisher = new gofshUtils.MasterFisher(lake, defs);
+  const fhirProcessor = new processor.FHIRProcessor(lake, fisher);
+
+  // Process the configuration
+  const configuration = fhirProcessor.processConfig(options.dependencies ?? []);
+
+  // Load dependencies, including those inferred from an IG file, and those given as input
```
ngfreiter, line 68: Why do I feel like I've seen these words somewhere before 🤔 .

ngfreiter (commented):
I'm not approving or requesting changes or anything, since it's just a feature branch you can merge whenever you want. Just had some comments you can address now if you want, or keep in mind to address later.

src/utils/RunSUSHI.js
```diff
@@ -11,6 +13,73 @@ const getRandomPun = utils.getRandomPun;
...
+    ? configuration.config.dependencies.map((dep) => `${dep.packageId}#${dep.version}`)
+    : [];
+  dependencies = sliceDependency(dependencies.join(','));
+  defs = await loadAndCleanDatabase(defs, dependencies);
+
+  // Process the FHIR to rules, and then export to FSH
+  const pkg = await gofshUtils.getResources(fhirProcessor, configuration);
+
+  // Return the string of FSH definitions
+  const fsh = new gofshExport.FSHExporter(pkg).apiExport('string'); // or use options.style
```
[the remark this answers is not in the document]
jafeltra: I don't think we'll ever want to split into separate "files" since the FSH will only ever go into the one editor on the LHS. I think I had left it while I was coming to that conclusions, so I can take it out now! Unless you think there would be a reason to split it.

src/utils/RunSUSHI.js
jafeltra: Ah yes, I may have copy and pasted 98% of that function. I'm adding a note at the top of this function to indicate that it's mainly pulled from `FhirToFsh`.

src/utils/RunSUSHI.js
jafeltra: Hm good point! This was really just here because I used so much of `FhirToFsh`, but since I'm not actually using that, I can simplify this for now and just `JSON.parse` each input resource.

src/utils/RunSUSHI.js
```diff
@@ -11,6 +13,73 @@ const getRandomPun = utils.getRandomPun;
...
+      }
+    }
+    if (gofshUtils.isProcessableContent(resource, location)) {
+      docs.push(new processor.WildFHIR(resource, location));
+    }
+  });
+
+  // Set up the FHIRProcessor
+  const lake = new processor.LakeOfFHIR(docs);
+  let defs = testDefs || new FHIRDefinitions(); // The tests pass in their own set of definitions to use
```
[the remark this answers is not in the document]
jafeltra: *sigh* Yes I completely agree. There should be a way to handle this. Also now that I've responded to your other comments, I feel like we really _should_ have a way to do this within the tests because if we do use `FhirToFsh` in the future, we can't use this approach. However, the problem I was having in the tests was that I can correctly mock out the `loadExternalDependencies` function, which gives back the test set of definitions and updates this variable `defs`. However, when doing that in the test, `fisher.external` isn't updated properly, but it is when the app is run. I'm not sure if maybe I need to somehow mock out `fisher` or `MasterFisher`, or if this is just relying on some object updating by reference that doesn't hold in the tests. If you have any ideas, let me know! I also think something similar might be happening in the tests for running SUSHI, but our test cases are so simple, I didn't notice until now.

src/utils/RunSUSHI.js
jafeltra: Okay actually it turns out that the test FSH that I have committed here is actually objects. I don't think the text we get from the editor will actually be objects, I think it will be strings, but do you think it's okay to keep this for now for our testing purposes and for alignment with `FhirToFsh`? Or do you think it's preferable to just stringify the temporary testing objects I have now?

src/utils/RunSUSHI.js
jafeltra: Note added in 487b9f1.

src/utils/RunSUSHI.js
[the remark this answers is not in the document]
jafeltra: Fixed in 487b9f1.

jafeltra:
Since GoFSH 1.0.1 was released with the optimizer changes, I removed the local version and included it as a regular dependency!

src/utils/RunSUSHI.js
jafeltra: I've stringified all the things! b2aa777

src/utils/RunSUSHI.js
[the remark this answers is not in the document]
jafeltra: You have helped me mock the right things and it works! Yay! b2aa777

jafeltra:
@ngfreiter all your comments are addressed now! Let me know if you see anything else, or if you're good to merge!

ngfreiter (approved):
Looks good!

julianxcarter (approved):
This looks good to me!

PR #38, opened 2021-03-04
Collapsible console

julianxcarter:
This PR address [CIMPL-683](https://standardhealthrecord.atlassian.net/browse/CIMPL-683), making the console expandable with an "Expand Console" button and also adding labels indicating how many warnings/errors resulted from the last run. I wanted the console to not make the actual editor page as a whole require any scrolling, so I've played around with some of the height values on the other components to get the height just right and not extend the content on the page to over 100% of the view screen.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,20 +14,24 @@ const useStyles = makeStyles((theme) => ({
     flexGrow: 1
   },
   itemTop: {
-    height: '75vh'
+    height: '85vh'
   },
-  itemBottom: {
-    height: '25vh'
+  console: {
+    height: '20vh'
   }
 }));
 
 const log = console.log; //eslint-disable-line no-unused-vars
 let consoleMessages = [];
 let errorAndWarningMessages = [];
+let errorCount = 0;
+let warningCount = 0;
 console.log = function getMessages(message) {
   consoleMessages.push(message);
   if (message && (message.startsWith('error') || message.startsWith('warn'))) {
     errorAndWarningMessages.push(message);
+    if (message.startsWith('error')) errorCount++;
+    else warningCount++;
   }
 };
 
@@ -56,6 +60,7 @@ export default function App(props) {
   const [initialText, setInitialText] = useState('Edit FSH here!');
   const [outputText, setOutputText] = useState('Your JSON Output Will Display Here: ');
   const [isOutputObject, setIsOutputObject] = useState(false);
+  const [expandConsole, setExpandConsole] = useState(false);
 
   useEffect(() => {
     async function waitForFSH() {
@@ -67,6 +72,8 @@ export default function App(props) {
   function resetLogMessages() {
     consoleMessages = [];
     errorAndWarningMessages = [];
+    errorCount = 0;
+    warningCount = 0;
   }
 
   function handleSUSHIControls(doRunSUSHI, sushiOutput, isObject) {
@@ -84,10 +91,10 @@ export default function App(props) {
       <TopBar />
       <SUSHIControls onClick={handleSUSHIControls} text={inputText} resetLogMessages={resetLogMessages} />
       <Grid className={classes.container} container>
-        <Grid className={classes.itemTop} item xs={6}>
+        <Grid style={{ height: expandConsole ? '55vh' : '85vh' }} item xs={6}>
           <CodeMirrorComponent value={inputText} initialText={initialText} updateTextValue={updateInputTextValue} />
         </Grid>
-        <Grid className={classes.itemTop} item xs={6}>
+        <Grid style={{ height: expandConsole ? '55vh' : '85vh' }} item xs={6}>
           <JSONOutput
             displaySUSHI={doRunSUSHI}
             text={outputText}
@@ -95,8 +102,15 @@ export default function App(props) {
             errorsAndWarnings={errorAndWarningMessages}
           />
         </Grid>
-        <Grid className={classes.itemBottom} item xs={12}>
-          <ConsoleComponent consoleMessages={consoleMessages} />
+        <Grid item xs={12}>
+          <ConsoleComponent
+            style={{ height: expandConsole ? '35vh' : '20vh' }}
+            consoleMessages={consoleMessages}
+            warningCount={warningCount}
+            errorCount={errorCount}
+            expandConsole={expandConsole}
+            setExpandConsole={setExpandConsole}
+          />
         </Grid>
       </Grid>
     </div>
--- a/src/App.js
+++ b/src/App.js
@@ -24,6 +24,8 @@ const useStyles = makeStyles((theme) => ({
 const log = console.log; //eslint-disable-line no-unused-vars
 let consoleMessages = [];
 let errorAndWarningMessages = [];
+let errorString = '';
+let warningString = '';
 let errorCount = 0;
 let warningCount = 0;
 console.log = function getMessages(message) {
@@ -32,6 +34,14 @@ console.log = function getMessages(message) {
     errorAndWarningMessages.push(message);
     if (message.startsWith('error')) errorCount++;
     else warningCount++;
+  }
+  if (errorCount > 0) {
+    errorString = `${errorCount} Error`;
+    if (errorCount !== 1) errorString += 's';
+  }
+  if (warningCount > 0) {
+    warningString = `${warningCount} Warning`;
+    if (warningCount !== 1) warningString += 's';
   }
 };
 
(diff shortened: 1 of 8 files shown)


cmoesel:
I was excited about this feature, so thought I'd give it a try.  It is super cool!  I really like it and I think the expand/collapse works really well!  I think we could release it as-is, but I have a suggestion or two to consider...

The console always shows the warnings, errors, and both associated icons:
image

I find that when there are errors, it doesn't really pop -- because only the number has changed.  And when there is success, it's fairly anticlimactic because nothing changes at all (unless there were previously errors).  So... what do you think about this?
* When you first load the page, there is no icon or warning/error count.
* When you run SUSHI:
  * If there are no errors or warnings: Green icon of some sort and "Success"
  * If there are only warnings: Warning icon and "X warnings"
  * If there are only errors: Error icon and "X errors"
  * If there are warnings and errors: Both icons and messages (like it is now)
* If you've already run SUSHI (so there is an icon and message showing already), as soon as the user hits "Run", clear the icon and message.  Since a run is not 0 seconds long, this should result in at least a flash before the new results are shown.

I feel like that would give it a bit more pop and noticeable feedback after running SUSHI.

cmoesel:
Also, it's just a bit more polished if we match pluralization to the number (i.e., 1 warning, not 1 warnings).

julianxcarter:
> I was excited about this feature, so thought I'd give it a try. It is super cool! I really like it and I think the expand/collapse works really well! I think we could release it as-is, but I have a suggestion or two to consider...
> 
> The console always shows the warnings, errors, and both associated icons:
> image
> 
> I find that when there are errors, it doesn't really pop -- because only the number has changed. And when there is success, it's fairly anticlimactic because nothing changes at all (unless there were previously errors). So... what do you think about this?
> 
> * When you first load the page, there is no icon or warning/error count.
> * When you run SUSHI:
>   
>   * If there are no errors or warnings: Green icon of some sort and "Success"
>   * If there are only warnings: Warning icon and "X warnings"
>   * If there are only errors: Error icon and "X errors"
>   * If there are warnings and errors: Both icons and messages (like it is now)
> * If you've already run SUSHI (so there is an icon and message showing already), as soon as the user hits "Run", clear the icon and message.  Since a run is not 0 seconds long, this should result in at least a flash before the new results are shown.
> 
> I feel like that would give it a bit more pop and noticeable feedback after running SUSHI.

I really like all of these suggestions, and I'll implement them right away!

cmoesel:
I just tried the new commit, @julianxcarter.  It is _awesome_!  Thank you.

jafeltra (commented):
This looks SO nice. I love the icons and they provide a really useful summary of what is happening. I just had a few small questions, mostly about layout. Let me know what you think!

- There's a blue border at the bottom of the console that appears when it is collapsed but disappears when it is expanded. I know we talked about colors, so is this intentional? If it is, is there a reason to have it disappear when the console is expanded?
- I'm not sure if this has to do with that blue border, but I have to scroll down slightly on the page to get the full text in the collapsed console in view (and see that blue border). Have you been able to get the collapsed console to fit on the screen without needing any scroll? I think it would feel really nice to have no scrollbar at all on the page as a whole and then only have the individual panes scroll as they need to.
- This is really small and I'm certainly no designer, but when I see the icons on the Warning and Error labels, I also kind of expect the arrows for collapsing to be on the left of the words as well. It somehow feels odd to me to have the icons on different sides of the words. Do you agree? Or do you feel its more natural to keep the arrows where they are?

julianxcarter:
I thought that all three of your suggestions (keeping the bottom border while the console was expanded, making sure that a page-wide scroll bar was never displayed, and also switching the icon on the "expand console" button) were excellent, and commit `b9ad505` should address them @jafeltra! 

There wasn't a real reason for the blue border disappearing when the console was expanded, as I had only added the border so that there wasn't whitespace under the console control bar and didn't think to add the border while the console was expanded as well.

jafeltra:
This is looking really good! I like the icon placement and I like the consistent blue border. I'm still seeing a slight scroll on the page so that the very bottom of the collapsed console grey bar and the blue border are cut off. Are you seeing anything similar? This is what my screen looks like: 
Screen Shot 2021-03-09 at 9 12 39 AM

I also notice that when I put it on my larger monitor, the pans don't seem to fill the page: 
Screen Shot 2021-03-09 at 9 12 54 AM

Space things are tough, so if the scroll bar issue in the first picture is tricky, I'm okay with leaving it for now - the scroll was never perfect to begin with. But I do think it would be good to fill the page when we are on a larger screen. I'm not really sure why that's happening though. I'm happy to help if I can though!

The only other thing besides those spacing issues that I think would be nice to include is a couple simple tests that the "Errors" and "Warnings" labels appear if we pass in a count in props.

julianxcarter:
@jafeltra I believe that [`b725540`](https://github.com/FSHSchool/FSHOnline/pull/38/commits/b7255406a017ccf56f9b5220dfacb65378a9c5bb) should address the spacing problems that we were running into, and also add in a test cases for the labels on the console bar!

src/App.js
```diff
@@ -80,23 +93,29 @@ export default function App(props) {
   }
 
   return (
-    <div className="root">
+    <div className="root" style={{ overflow: 'hidden', height: '100vh' }}>
       <TopBar />
       <SUSHIControls onClick={handleSUSHIControls} text={inputText} resetLogMessages={resetLogMessages} />
       <Grid className={classes.container} container>
-        <Grid className={classes.itemTop} item xs={6}>
+        <Grid style={{ height: expandConsole ? '55vh' : '85.3vh' }} item xs={6}>
```
jafeltra, line 100: Is there a reason to move these from the classes to a `style` tag? I think generally we want to keep styles in the classes if we can.

Similarly for the `Grid` below this one.

src/components/ConsoleComponent.js
```diff
@@ -24,9 +46,33 @@ const theme = createMuiTheme({
...
+  const toggleExpandConsole = () => {
+    props.setExpandConsole(!props.expandConsole);
+  };
+
   return (
     <ThemeProvider theme={theme}>
-      <Box className={classes.box} overflow="scroll">
+      <Box
+        className={classes.consoleControls}
+        style={{ borderBottom: !props.expandConsole ? '6px solid #2c4f85' : '' }}
```
jafeltra, line 57: It looks like the border that's on the collapsed console is 4px, as specified by the `box` class above. Is there a reason for changing between the two sizes? If the border is causing headaches, I think we could remove it if we wanted.

jafeltra (commented):
So I really don't want to do this, but it doesn't seem quite right still. I had a few questions about styling inline below. However, high level, now when I put it on my larger screen, the blue banner at the top gets expanded, rather than staying at a fixed height like it used to.
Screen Shot 2021-03-10 at 4 02 54 PM

I'm not entirely sure that I love either this version or the previous version, but if I had to pick one, it might be the first one with the white space at the bottom but with the fixed blue bar at the top. But I also recognize that there's been a lot of time put into this, so I'm not sure when we draw the line on this. I'm not sure if you @julianxcarter or @cmoesel have any thoughts on that.

src/App.js
julianxcarter: I was using the `style` tag because I wanted to make the height of many of the components dependent on the `expandConsole` property that's created in the main app, which I didn't think I could do in the classes since that variable arrives as part of the `props` arguments in the ConsoleComponent. I think I could just move the styles object down a level into the main `App` function and that should be a workaround.

Evidently that may not be what we want to do anyway, I'm beginning to think that these ternary statements aren't the best way to approach resizing the components on the screen 😕.

src/components/ConsoleComponent.js
julianxcarter: I was just going for a height balance on the screen to prevent scroll! 

I definitely think that the border is easily salvageable if we really want that in there, but I'm at a loss as to how to we'd prevent the scroll on the entirety of the page without making these components resize in a weird way like they're doing now. I'd love to brainstorm with you on this separately to see if we can come up with a workaround!

jafeltra: Yeah definitely! I had one idea about a slightly different approach to fixing the heights, but I tried once quickly yesterday and wasn't successful. I can give it one more shot or work with you to try for a bit!

jafeltra:
@julianxcarter I pushed up the changes we just talked about for the heights. Let me know if the diffs make sense and if things still look good on your screens!

jafeltra (approved):
This looks great! Thanks for so much back and forth on the spacing! It looks great now, and is a much more useful console!

cmoesel (approved):
This is super cool!  I like it!  My only suggestion would be that perhaps if there is an error, we should automagically expand the console (so it's obvious to the user, but also because usually the user will want to expand the console anyway to see the error details).  But I think we might have discussed this already and decided against it for at least this PR.

But when @jafeltra removes the error message from the RHS, I think maybe we'll want to consider this idea again...

PR #39, opened 2021-03-10
Initial GoFSH editor

jafeltra:
This PR replaces the current RHS of FSH Online with an editor that can be used to for GoFSH inputs.

Currently, when you type in FSH on the LHS and run SUSHI, the editor on the RHS will populate with the first definition in the package. There is a very simple file navigator on the right that will let you switch to any other definitions that were in the package. Clicking a different definition should update the text in the editor, but the majority of the styling and organization for the file list will be part of a different PR (because it is a separate task). Additionally, you can edit the text in the RHS editor and then run GoFSH. The LHS FSH editor should populate with the resulting FSH.

This PR also updates the editor to use a placeholder plugin to display the initial text in the editors. It also removes the "Errors and Warnings" summary that currently displays on the RHS of FSH Online. There wasn't an obvious place I could think to display those, and the changes Julian made to the console should make it more obvious when there are errors or warnings, so I took them out of the RHS.

Note: There are still some components and files that have very "SUSHI" specific names, but since changing those would make the diffs difficult to understand, I'm planning on updating that in it's own PR at some point.
Another note: There is one test skipped that was supposed to be for testing that text is populated to the JSON editor on the RHS. However, the text that gets populated in the test is not right and I haven't been able to figure the right way to test that, but I thought it was more important to put up this PR.

### DIFF
--- a/src/components/SUSHIControls.js
+++ b/src/components/SUSHIControls.js
@@ -186,7 +186,7 @@ export default function SUSHIControls(props) {
         <Button className={classes.secondaryButton} onClick={handleOpenShare}>
           Share
         </Button>
-        <Button className={classes.button} onClick={handleGoFSHClick}>
+        <Button className={classes.button} onClick={handleGoFSHClick} testid="GoFSH-button">
           Run GoFSH
         </Button>
         <Dialog open={openConfig} onClose={handleCloseConfig} aria-labelledby="form-dialog-title">
--- a/src/components/SUSHIControls.js
+++ b/src/components/SUSHIControls.js
@@ -89,7 +89,7 @@ export default function SUSHIControls(props) {
   };
 
   const handleOpenShare = async () => {
-    const encoded = deflateSync(props.text).toString('base64');
+    const encoded = deflateSync(props.fshText).toString('base64');
     const longLink = `https://fshschool.org/FSHOnline/#/share/${encoded}`;
     const bitlyLink = await generateLink(longLink);
     if (bitlyLink.errorNeeded === true) {
@@ -139,11 +139,11 @@ export default function SUSHIControls(props) {
   //Sets the doRunSUSHI to true
   async function handleRunClick() {
     props.resetLogMessages();
-    props.onClick(true, 'Loading...', false);
+    props.onClick(true, 'Loading...', false, true);
     let isObject = true;
     const dependencyArr = sliceDependency(dependencies);
     const config = { canonical, version, FSHOnly: true, fhirVersion: ['4.0.1'] };
-    const outPackage = await runSUSHI(props.text, config, dependencyArr);
+    const outPackage = await runSUSHI(props.fshText, config, dependencyArr);
     let jsonOutput = JSON.stringify(outPackage, replacer, 2);
     if (outPackage && outPackage.codeSystems) {
       if (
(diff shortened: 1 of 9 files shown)


src/components/SUSHIControls.js
```diff
@@ -161,69 +161,17 @@ export default function SUSHIControls(props) {
...
-        {
-          uri: 'http://hl7.org/fhir/us/mcode/ImplementationGuide/hl7.fhir.us.mcode',
-          packageId: 'hl7.fhir.us.mcode',
-          version: '1.0.0'
-        }
-      ]
-    };
+    props.onGoFSHClick('Loading...', true);
+    const gofshInputStrings = [];
+    props.gofshText.forEach((def) => gofshInputStrings.push(JSON.stringify(def)));
```
ngfreiter, line 170: Is this supposed to work in the case that you just start by pasting JSON on the RHS and try to run GoFSH? I hit a bug in that case, and it looks like it is because `gofshText` is initialized to `''`, so `forEach` does not exist on it. If you run SUSHI first from FSH on the LHS, then things work out because `gofshText` becomes an array, so now you can run GoFSH.

ngfreiter (changes requested):
I noticed one issue when trying to directly convert JSON to FSH without first going from FSH to JSON. Is this supposed to work now? Or are you planning on making this work along when you add more to the file tree?

src/components/SUSHIControls.js
jafeltra: Oh good call! I think that should work as part of this, but I definitely hadn't tried it, and that sounds like it makes sense that it breaks. Thanks for finding it!

jafeltra: I had too restrictive of a check for not updating the text in the editor. It was assuming you'd only be editing after a SUSHI package had come back and was displayed. e10a50f gets rid of that condition and should fix this bug.

ngfreiter (approved):
Looks good to me now!

julianxcarter (commented):
This looks great to me overall! I did notice a bug with the JSONOutput component that should be shown in the screen recording below. When you attempt to paste in a FHIR definition that exceeds the line count of the component, the code mirror component initially doesn't display the resource and returns an `Uncaught Error` to the console. When you click on the JSON input though, everything appears as normal.

I'm not entirely sure why that happens in the RHS editor, because the LHS editor handles large pastes from the clipboard without any issues and they're both using the same component.

https://user-images.githubusercontent.com/53443041/111159643-b5a0a400-856f-11eb-9f7d-50f26e08511d.mov

jafeltra:
Good find @julianxcarter! I think I must have been using too simple of examples and never saw this. I can't exactly tell you why it wasn't working, but the main difference between the editor component for FSH vs JSON was how the `initialText` and `value` props were tied together. The JSON one always updated both and the FSH one only used the `initialText` as truly initial text. I had started using the `initialText` prop in that way on the file tree branch I'm on, so I decided to move that over here and it seems to have resolved the problem!

I'm also not entirely sure how to test this case since I think it would be similar to the test that I currently have skipped and still am not sure how to get working in properly 😕

julianxcarter:
Hmmm maybe the CodeMirror component glitched out when its initial text value changed after being rendered. In any case, your most recent commit fixed things for me!

src/tests/components/JSONOutput.test.js
```diff
@@ -15,62 +24,65 @@ afterEach(() => {
...
 
-  expect(resultsElement).toBeInTheDocument(); // isObject is true so results are printed
-  expect(errorsElement).not.toBeInTheDocument(); // No errors
+  expect(textElement).toBeInTheDocument(); // Mimics the 'loading...' case
+  expect(textElement.parentNode.textContent).toEqual('Loading...');
 });
 
-it('Renders error messages if present', () => {
+// TODO: CodeMirrorComponent doesn't get the package text properly - not sure why
+it.skip('Renders with the first profile when text is an object (SUSHI Package)', async () => {
```
julianxcarter, line 54: I played around with this skipped test case for a little bit and was stumped on it too. It's really not making sense to me that the `getByText` function is not finding the content there but working just fine in the test right above it 😕. We can debug this later though and everything is working great, so I'm still gonna go ahead and approve. This feature is looking really awesome so far, nice work!

julianxcarter (approved):

src/tests/components/JSONOutput.test.js
jafeltra: Yeah, the weird thing to me is that you can see the output when the test fails and it looks like there's only a `}` in the editor, which is really odd. As long as you think we're okay holding off on this test for now, I'm good! I'd love some help eventually sorting out the test when we do fix it.

PR #40, opened 2021-03-16
Updated JSON Definition File Tree

jafeltra:
This PR updates the file tree that is rendered on the RHS editor.

A few features of the file tree:
- If the JSON entered for that definition is valid, a check mark is displayed next to its name
- If the JSON is invalid, an X is displayed next to it and the text is colored red and bolded (I don't actually think the bolding is necessary since there is a non-color based visual with the X and I didn't mind it, so I left it for others' opinions)
- Choosing a different definition switches the definition loaded in the editor
- Changing the `id` property in the JSON updates the text in the file tree
- Changing the `resourceType` will re-sort the definition. If it is anything besides "StructureDefinition", "ValueSet", or "CodeSystem", it is sorted under "Instances".

A few implementation/code notes:
- I updated the type of `fhirDefinitions` in the `JSONOutput` component to be a string so that any JSON value is stored, not just valid JSON, letting you keep invalid JSON in one definition while viewing another
- The grey background color of selected definitions doesn't fill the name if the name is long enough to require a horizontal scroll. I couldn't figure out how to do this, so if anyone has any ideas, let me know.
- There isn't a test for typing in invalid JSON and testing that the `listItemError` class is added because I can't figure out how to type directly in the editor within the test. Again, if anyone has any ideas, I'm all for them.
- I did set a `style={{ height: '75vh' }}`, which I think will need to change with the changes to the collapsable console. I'm not sure if now is the right time to rebase the `gofsh` branch (and then this one) to have those changes properly included.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1525,6 +1525,14 @@
         "react-transition-group": "^4.4.0"
       }
     },
+    "@material-ui/icons": {
+      "version": "4.11.2",
+      "resolved": "https://registry.npmjs.org/@material-ui/icons/-/icons-4.11.2.tgz",
+      "integrity": "sha512-fQNsKX2TxBmqIGJCSi3tGTO/gZ+eJgWmMJkgDiOfyNaunNaxcklJQFaFogYcFl0qFuaEz1qaXYXboa/bUXVSOQ==",
+      "requires": {
+        "@babel/runtime": "^7.4.4"
+      }
+    },
     "@material-ui/styles": {
       "version": "4.10.0",
       "resolved": "https://registry.npmjs.org/@material-ui/styles/-/styles-4.10.0.tgz",
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@material-ui/core": "^4.10.2",
+    "@material-ui/icons": "^4.11.2",
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
--- a/src/App.js
+++ b/src/App.js
@@ -53,7 +53,7 @@ export default function App(props) {
   const text64 = props.match.params;
   const [doRunSUSHI, setDoRunSUSHI] = useState(false);
   const [inputFSHText, setInputFSHText] = useState('');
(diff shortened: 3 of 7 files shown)


ngfreiter (commented):
Overall this looks good! I have a few comments below.

I'm no designer, but for me I don't think the check mark is really necessary. I don't think we need to signal when the JSON is valid, and can get away with just signaling when it is not valid (with the red x and so on).

I've noticed that the `Loading` text for GoFSH is red, is that intentional:
image

I've found an issue where if I press the "Run SUSHI" and then "Run GoFSH" buttons in quick succession, I get the following crash:
image

I don't think we have to do anything smart in this case I would be fine with the "Run GoFSH" just not working, or with the "Run GoFSH" running after the "Run SUSHI" has finished. Whatever you think works more smoothly to be honest. I just think we want to avoid the crash, since it's a plausible thing to happen.

jafeltra:
@ngfreiter I think all three of your comments should be resolved by the last two commits (e3d95d7 and 379acfa). I got rid of the check mark icon and replaced it with some blank space because I thought it looked weird to have the invalid definition names get shifted to the right to add the X icon. I also updated both editors to display "Loading..." in their placeholder text. I think this is probably how it should have been the whole time. It also let me consistently return either a string (for FSH) or an array of strings (for GoFSH), which also eliminated the crash. Let me know what you think!

jafeltra:
Actually.. the placeholder updates might not be perfect. It doesn't look like the text ever gets cleared out to show the "Loading..." placeholder text if there is already text in the editor. I might need to look at that tomorrow morning.

jafeltra:
Julia - I regret to inform you, but I have found a bug on your branch. To reproduce it, remove any `id` field from a definition. Disaster! 

image

cmoesel:
> Julia - I regret to inform you, but I have found a bug on your branch. To reproduce it, remove any `id` field from a definition. Disaster!
> 
> image

Oh man.  Julia is going to hate you now for finding that.

jafeltra:
Never fear @jafeltra and @cmoesel, I've discovered a fix! d009095 should do the trick! Thanks for uncovering this issue!

ngfreiter (approved):
This looks good to me, especially now that Julia fixed that bug that Julia found.

julianxcarter (approved):
This is a really cool feature, and it looks great and works well for me. I'm approving, awesome work! @jafeltra

PR #41, opened 2021-03-22
Add Definition and Delete Definition buttons

jafeltra:
This PR adds a "Add definition" button to the file tree on the RHS. This lets you add a blank definition and open a blank editor to type the definition into. As you add text to the definition, the definition should be sorted into the correct category in the file tree and the id should populate.

This also adds a delete button so you can remove a full definition. Deleting the definition will remove it from the file tree and will remove it from the set of definitions that GoFSH is run on if you use the "Run GoFSH" button.

### DIFF
--- a/src/components/JSONOutput.js
+++ b/src/components/JSONOutput.js
@@ -176,19 +176,34 @@ export default function JSONOutput(props) {
 
   const handleCloseAndDelete = (index) => {
     const updatedDefs = [...fhirDefinitions];
-    updatedDefs.splice(index, 1);
+    updatedDefs.splice(index, 1); // Remove definition to be deleted
     if (updatedDefs.length === 0) {
+      // Ensure list is not empty so there is always a definition in the file tree
       updatedDefs.push({ resourceType: null, id: 'Untitled', def: null });
     }
     setFhirDefinitions(updatedDefs);
-    setCurrentDef(0);
-    if (defsWithErrors.includes(index)) {
-      const newErrors = [...defsWithErrors];
-      newErrors.splice(defsWithErrors.indexOf(index), 1);
-      setDefsWithErrors(newErrors);
+
+    // If we deleted the currently viewed definition or any before it, update it
+    if (index <= currentDef) {
+      let newCurrentDef = 0;
+      if (index < currentDef) {
+        newCurrentDef = currentDef - 1;
+      }
+      setCurrentDef(newCurrentDef);
+      const newCurrentDefText = updatedDefs.length > 0 ? updatedDefs[newCurrentDef].def : null;
+      setInitialText(newCurrentDefText);
     }
-    const newCurrentDef = updatedDefs.length > 0 ? updatedDefs[0].def : null;
-    setInitialText(newCurrentDef);
+
+    // Update error tracking
+    const newErrors = [...defsWithErrors];
+    if (defsWithErrors.includes(index)) {
+      // If the definition had an error, remove it from the list
+      newErrors.splice(defsWithErrors.indexOf(index), 1);
+    }
+    // Shift any indices that came after the deleted definition to properly track errors
+    const shiftedErrors = newErrors.map((i) => (i < index ? i : i - 1));
+    setDefsWithErrors(shiftedErrors);
+
     setOpenDeleteConfirmation(false);
     props.updateTextValue(updatedDefs);
   };
--- a/src/components/JSONOutput.js
+++ b/src/components/JSONOutput.js
@@ -84,7 +84,7 @@ const getIterablePackage = (defsPackage) => {
 export default function JSONOutput(props) {
   const classes = useStyles();
   const [initialText, setInitialText] = useState('');
-  const [fhirDefinitions, setFhirDefinitions] = useState([{}]);
+  const [fhirDefinitions, setFhirDefinitions] = useState([{ resourceType: null, id: 'Untitled', def: null }]);
   const { setIsOutputObject, updateTextValue: propsUpdateText } = props;
   const [currentDef, setCurrentDef] = useState(0);
   const [defsWithErrors, setDefsWithErrors] = useState([]);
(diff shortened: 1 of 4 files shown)


ngfreiter (changes requested):
This looks good. The only thing I've noticed is that the `Add Definition` button doesn't stretch across the entire sidebar in the same way the definitions do: 
image

I think it would make more sense to have that button be the same width as the definitions below.

cmoesel:
This is very nice.  I really like it.  That said, I found a few issues in playing with it.  Some of which may be related to each other.  I've done screen captures for these -- so hopefully they come through.

1. A blank RHS is not noted as an error at first, but if you type in JSON and then delete it all (back to blank state), it is now an error.  I guess I'm not sure if that's good or bad.
FSH Online-20210324131428

2. If you add two (or more) untitled resources (for a total of three or more), then click on #1 and enter JSON, then click on #2 and enter JSON, it works. But then if you click on #3, the editor still shows #2 (no matter how many times you click).  The only way to get the blank editor is to click back on #2 again and then click on #3 again.
FSH Online-20210324132003

3. This is kind of silly, but if the first thing you do is delete the Untitled artifact (so there are no files on the RHS), then the editor is still active --- so if you type in a resource, as soon as the resource is valid JSON, the editor goes blank and the resource appears in the file tree.  So then you need to click on the resource in the file tree to continue editing.
FSH Online-20210324132623

4. I can make it crash!  Start fresh and type a FSH profile on the LHS.  Then run SUSHI.  Then "Add Definition" on the RHS.  Then run SUSHI again.   CRRRRRAAAAAASSSSHHHHHH!
FSH Online-20210324133106

cmoesel (changes requested):
Oops.  Forgot to request changes when I commented on the bugs I found.

jafeltra:
Thanks for finding those bugs @cmoesel. A quick summary of the fixes:

1. I agree that the inconsistency was weird and it felt odd to have a blank editor be invalid, so I just had an empty string (which is technically invalid JSON) be treated as an empty JSON object.
2. This was a bummer because I had tried something similar, noticed similar weirdness, and thought I fixed it, but apparently it only worked with 2 new definitions and I never tried with more. The root of this problem is that the CodeMirror component is using the `initialText` prop to rerender and clear out the editor. But when you add the group of definitions like you did, the `initialText` is always `null` and so switching to the other untitled editor doesn't clear the text that's present properly. Or at least that's what I think is happening. In order to get around that, I reset the `initialText` to be something other than `null` when you first start typing. This lets the text in the editor get managed by CodeMirror as you type, but it lets the `initialText` prop on the editor register a change when you swap editors. This is definitely a little odd, so let me know if you feel too iffy on this.
3. I'm not entirely sure why this happened, but I decided it was a little weird to have that empty list of files since we start the list out with a blank definition on load. So I decided to make sure we add back a blank definition if we have 0 definitions listed in the tree. So I think that resolves this bug.
4. The crash was happening because the "currently selected" definition index wasn't getting reset when we load in new definitions from SUSHI. Resetting that back to 0 selects the definition as expected and doesn't crash.

Bonus: I moved and renamed the "Share" button to make it more clear that we are sharing FSH, not JSON, like we talked about this morning.

cmoesel (changes requested):
So this is looking good!  But I heard you like weird bugs.  So I have just one more for you. To reproduce:

* Refresh your page
* Add 3 more resources (for total of four)
* Edit them one at a time, top to bottom
* Go to 2nd item and create syntax error
* Delete 1st item

Result: Original 2nd item (now 1st item) will still be in red (good) but the original 3rd item (now 2nd item) will become red even though it has no errors (bad).  Merely clicking on it clears the error indicator.

FSH Online-20210324221050

jafeltra:
Good bug finding @cmoesel! 6259a16 should fix it, along with one other tweak. The problem was that the definitions that have errors are kept track of through a separate array of indices, so that array wasn't being updated relative to the changing source array. That commit adjusts the indices to update any later occurring definitions. While looking at the function that handles deleting again, I also adjusted which definition is selected and displayed after deleting as well. Previously, any time a definition was deleted, the first one was reselected. Now it will only reselect the first one if you delete the one you are currently viewing and will shift to keep the same definition selected if you delete one that comes before it.

cmoesel (changes requested):
It's working nice!  I found one behavior that feels a little odd to me, but it's not necessarily wrong (so if it's not easily fixed just say so).  Basically, if the 3rd (or 4th, etc) resource in a list is untitled, it will stay in its position as you edit until you have a parseable resourceType and id.  This makes sense.  But if you enter a blank id (`""`), it's name will stay "Untitled" but its position will hop up to the top of the list.  It's jarring to have the position change even though the name did not.
FSH Online-20210325124911

This is kind of an edge case, but I wonder if it will actually be aggravated when this is combined w/ auto-closing `""` since whenever you type an `id` it will always start with an empty string (and thus shoot to the top of the list until you type something into the string).

jafeltra:
@cmoesel I believe this should be resolved now. I think that the filtering by id was changed whether the id was an empty string or just not defined. I resolved it by sorting all definitions that didn't have an `id` by the "Untitled" name, which is what they do when you first add them using the "Add Definition" button. I think the only potentially odd result of this is that they do get sorted alphabetically still, so you could end up with a list that looks like this: 

image

I think that is what I would expect, but want to make sure you agree.

I also discovered a bug where adding a bunch of blank definitions using the "Add Definition" button and sending them to GoFSH caused the app to crash. This was because `null` can be successfully parsed by `JSON.parse()`, so we didn't get the nice error message before sending definitions to GoFSH, but then GoFSH couldn't read `resourceType` from `null`, as expected. This now filters out any `null` or `undefined` definitions that we try to send to GoFSH.

cmoesel (approved):
Splendid!  Thanks for fixing all the silly probably-will-never-happen-in-the-real-world things for me!

ngfreiter (approved):
Chris took all the good breaking, so I'm unable to break it. Looks good to me!

PR #42, opened 2021-03-24
Code Fold and Auto Close Bracket plugins

jafeltra:
This PR adds two plugins to both editors:
- Auto close brackets will automatically close `()`, `[]`, `{},` `''`, `""`
- Code folding will let you collapse different brackets in the JSON editor. There is an arrow in the gutter that lets you collapse/expand sections similar to VS Code. Also, the example in the documentation also added the collapse/expand to the `Ctrl-Q` keyboard shortcut, so I included that as well. When a section is collapsed, it can also be expanded by clicking the arrow that appears between the collapsed brackets.

Since this just adds existing code mirror plugins, I couldn't think of any tests to add.

### DIFF
--- a/src/components/CodeMirrorComponent.js
+++ b/src/components/CodeMirrorComponent.js
@@ -10,6 +10,7 @@ import 'codemirror/addon/fold/foldgutter.css';
 require('codemirror/addon/mode/simple');
 require('codemirror/addon/edit/closebrackets');
 require('codemirror/addon/display/placeholder');
+require('codemirror/addon/comment/comment');
 require('codemirror/addon/fold/foldgutter');
 require('codemirror/addon/fold/brace-fold');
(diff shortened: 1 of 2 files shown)


jafeltra:
I just added in one other plugin that I thought would be nice, but I can take it out if others are less sure since it wasn't part of the task. 
- The toggleComment plugin lets you toggle a line to be commented or uncommented, similar to VS Code. I added keybindings for `Cmd-/` and `Ctrl-/` for this because I wanted it to work as expected for Mac and Windows. Since JSON doesn't support comments, this only works in the FSH editor right now.

ngfreiter (approved):
This looks good to me! Thanks for adding the comments shortcut too, that is very useful.

cmoesel (approved):
Yep, looks good to me too!

PR #43, opened 2021-03-24
GoFSH Console Improvements

ngfreiter:
This improves the GoFSH logging in a few ways:

* Adds a cool box
* Fixes a bug which would cause GoFSH logs not to overwrite SUSHI ones
* Fixes a bug which would cause GoFSH to count errors from previous runs

### DIFF
--- a/src/utils/RunSUSHI.js
+++ b/src/utils/RunSUSHI.js
@@ -65,6 +65,7 @@ export async function runGoFSH(input, options) {
 
   // Return the string of FSH definitions
   const fsh = new gofshExport.FSHExporter(pkg).apiExport('string');
+  logger.info('Done converting definitions');
   printGoFSHresults(pkg);
   return fsh;
 }
--- a/src/utils/RunSUSHI.js
+++ b/src/utils/RunSUSHI.js
@@ -187,6 +187,6 @@ function printGoFSHresults(pkg) {
     '╚'  + '═════════════════════════════════════════════════════════════════' +'╝'
   ];
 
-  console.log();
+  console.log(' ');
   results.forEach((r) => console.log(r));
 }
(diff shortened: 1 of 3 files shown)


cmoesel (approved):
This looks good to me!  I couldn't break it!

I expect this might cause some extra work / conflicts, however, when we try to rebase/merge w/ master and the collapsible console there.

src/utils/RunSUSHI.js
```diff
@@ -149,3 +153,40 @@ function printResults(pkg) {
...
+    '║' + ' ├──────────────┼───────────────┼──────────────┼───────────────┤ ' + '║',
+    '║' + ` │ ${ instNum } │ ${  invNum  } │ ${ mapNum  } │ ${creatrStat} │ ` + '║',
+    '║' + ' ╰──────────────┴───────────────┴──────────────┴───────────────╯ ' + '║',
+    '║' + '                                                                 ' + '║',
+    '╠'  + '═════════════════════════════════════════════════════════════════' +'╣',
+    '║' + ` ${aWittyMessageInvolvingABadFishPun } ${errNumMsg} ${wrnNumMsg} ` + '║',
+    '╚'  + '═════════════════════════════════════════════════════════════════' +'╝'
+  ];
+
+  console.log();
```
jafeltra, line 190: I think this is trying to add a blank line before the box, but it doesn't look like it's getting added. The SUSHI one logs a `' '`, which I think should work here, if that was the intension.

jafeltra (commented):
This looks great! I had one super small comment inline. And I have one more small question not about a line of code:

- The last log message we get is "Optimizing FSH definitions to follow best practices..." When run in the terminal, we get another couple output messages about what files we wrote. Obviously we're not writing any files here, but do you think it's worth while to add one more message here saying something like "Output optimized definitions" so we know we reached the end? Or do you think the pretty box accomplishes that? This is truly an opinion question.

src/utils/RunSUSHI.js
ngfreiter: Fixed in https://github.com/FSHSchool/FSHOnline/pull/43/commits/dacbc2f697103e8ee0f3759fe0f6951a057de2fb

jafeltra (approved):
Wonderful.

PR #44, opened 2021-03-29
Update gofsh Branch

jafeltra:
This PR updates the `gofsh` branch with the current version of the `master` branch. The only feature that the branch was missing is the collapsible console. However, since that feature updated the way we size the various components, reviewers should check that the spacing for both editors and the file tree still looks good with the console collapsed and expanded.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -11,23 +11,54 @@ import SUSHIControls from './components/SUSHIControls';
 
 const useStyles = makeStyles((theme) => ({
   container: {
+    height: '100%',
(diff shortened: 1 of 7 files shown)


ngfreiter (changes requested):
This looks good to me, only minor issue I can see is if you have a lot of definitions, they overlap the console:
image

ngfreiter (approved):
I am sold!

julianxcarter (approved):
This looks great to me!

PR #45, opened 2021-04-01
Component rename and clean up

jafeltra:
This PR doesn't add any new functionality. It just updates the name of some components, files, and functions so they match up better with their current functionality. For example, the RunSUSHI file does more than just run SUSHI now. I also cleaned up some of the props and simplified some of the variables used to determine when updates needed to happen to the RHS. And I added a brief sentence about GoFSH in the readme.


cmoesel (approved):
This all looks good to me and the interaction of the app seems to still work, so that's good too!  At first I thought the "Share" function was broken, but I just needed to setup my REACT_APP_BITLY_KEY.

So... nice job!

README.md
```diff
@@ -1,6 +1,6 @@
 # FSH Online
 
-FSH Online is a web application for authoring [FHIR Shorthand (FSH)](https://build.fhir.org/ig/HL7/fhir-shorthand/) and running the [SUSHI](https://github.com/FHIR/sushi) compiler on the authored FSH directly in a web browser. It is available on https://fshschool.org/FSHOnline/.
+FSH Online is a web application for authoring [FHIR Shorthand (FSH)](https://build.fhir.org/ig/HL7/fhir-shorthand/) and running the [SUSHI](https://github.com/FHIR/sushi) compiler on the authored FSH directly in a web browser. It also runs the [GoFSH](https://github.com/FHIR/GoFSH) decompiler to translate FHIR definitions into FSH. It is available on https://fshschool.org/FSHOnline/.
 
 NOTE: This application is an early alpha release and is designed to be used for initial exploration.
```
ngfreiter, line 5: Not saying we need to remove this here, but I am wondering what has to happen for us to no longer consider this "alpha"?

ngfreiter (approved):
Looks good to me, I had one question, but it isn't something I imagined you addressing in this PR, more just a general thing that made me wonder.

README.md
jafeltra: Ah that's a good question. I'm not entirely sure either. There's no real versioning so I think this reference is the only place we call it an alpha. I'd feel comfortable taking that line out when we release the `gofsh` branch, but we can also discuss as a team!

PR #46, opened 2021-04-06
Canonical and Version configuration updates

jafeltra:
This PR updates how the Configuration modals works with GoFSH and SUSHI so that we can successfully "round trip" within FSH Online.

When no values are set, we leave the fields blank and indicate the default values that SUSHI will use. If there are no values set when SUSHI runs, it will use the default `example.org` and `1.0.0` values. If there are no values when GoFSH is run, GoFSH will infer the values from the definitions, and then set the canonical and version value in modal, which will be used by SUSHI the next time SUSHI is run, completing the round trip.

If there are values set in modal when running SUSHI, SUSHI will use those. If there are values set in the modal when running GoFSH, GoFSH will use those as the configuration and set caret value rules for the actual canonical and version values in the definitions. It will not update the modal with the configuration it determines from that set of definitions, since the modal already has values and will not be overwritten.

### DIFF
--- a/src/components/FSHControls.js
+++ b/src/components/FSHControls.js
@@ -175,7 +175,8 @@ export default function FSHControls(props) {
       const igResource = {
         resourceType: 'ImplementationGuide',
         fhirVersion: ['4.0.1'],
-        ...(canonical && { url: canonical }),
+        id: '1',
+        ...(canonical && { url: `${canonical}/ImplementationGuide/1` }),
         ...(version && { version: version })
       };
       gofshInputStrings.push(JSON.stringify(igResource, null, 2));
--- a/src/components/FSHControls.js
+++ b/src/components/FSHControls.js
@@ -75,8 +75,8 @@ export default function FSHControls(props) {
   const [openShareError, setOpenShareError] = useState(false);
   const [link, setLink] = useState();
   const [{ copied, copyButton }, setCopied] = useState({ copied: false, copyButton: 'Copy to Clipboard' });
-  const [canonical, setCanonical] = useState('http://example.org');
-  const [version, setVersion] = useState('1.0.0');
+  const [canonical, setCanonical] = useState('');
+  const [version, setVersion] = useState('');
   const [dependencies, setDependencies] = useState('');
 
   const handleOpenConfig = () => {
@@ -139,7 +139,12 @@ export default function FSHControls(props) {
     props.resetLogMessages();
     props.onSUSHIClick(true, [''], true);
     const dependencyArr = sliceDependency(dependencies);
-    const config = { canonical, version, FSHOnly: true, fhirVersion: ['4.0.1'] };
+    const config = {
+      canonical: canonical ? canonical : 'http://example.org',
+      version: version ? version : '1.0.0',
+      FSHOnly: true,
+      fhirVersion: ['4.0.1']
+    };
     const outPackage = await runSUSHI(props.fshText, config, dependencyArr);
     let jsonOutput = JSON.stringify(outPackage, replacer, 2);
     if (outPackage && outPackage.codeSystems) {
@@ -164,9 +169,23 @@ export default function FSHControls(props) {
     props.onGoFSHClick('', true);
     const gofshInputStrings = props.gofshText.map((def) => def.def).filter((d) => d);
     const parsedDependencies = dependencies === '' ? [] : dependencies.split(',');
+
+    // Create small ImplementationGuide resource to send canonical and version information
+    if (canonical || version) {
+      const igResource = {
+        resourceType: 'ImplementationGuide',
+        fhirVersion: ['4.0.1'],
+        ...(canonical && { url: canonical }),
+        ...(version && { version: version })
+      };
+      gofshInputStrings.push(JSON.stringify(igResource, null, 2));
+    }
+
     const options = { dependencies: parsedDependencies };
-    const fsh = await runGoFSH(gofshInputStrings, options);
+    const { fsh, config } = await runGoFSH(gofshInputStrings, options);
     props.onGoFSHClick(fsh, false);
+    if (canonical === '' && config.canonical) setCanonical(config.canonical);
+    if (version === '' && config.version) setVersion(config.version);
   }
 
   return (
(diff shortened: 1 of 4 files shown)


src/components/FSHControls.js
```diff
@@ -164,9 +169,23 @@ export default function FSHControls(props) {
     props.onGoFSHClick('', true);
     const gofshInputStrings = props.gofshText.map((def) => def.def).filter((d) => d);
     const parsedDependencies = dependencies === '' ? [] : dependencies.split(',');
+
+    // Create small ImplementationGuide resource to send canonical and version information
+    if (canonical || version) {
+      const igResource = {
+        resourceType: 'ImplementationGuide',
+        fhirVersion: ['4.0.1'],
+        ...(canonical && { url: canonical }),
```
cmoesel, line 178: Technically, I think that the IG `url` is not the canonical, but is _based_ on the canonical.  So it really should be something more like: ``${canonical}/ImplementationGuide/1`` (which assumes the IG JSON has id `1` - which you might as well also add).

cmoesel (changes requested):
This _mostly_ works really well and I like it!  The only issue I ran into is that if you populate _only_ the version in the configuration and then run GoFSH, you get this error:
> error Could not process StructureDefinition at Input_0: config is undefined

It works fine if you populate _only_ the Canonical.  But _only_ the Version seems to be a problem.

I also submitted one small comment about the generated IG URL.

src/components/FSHControls.js
jafeltra: I should have thought about this, so thanks for noticing! I'm changing it to be the pattern you described, which is what SUSHI uses to build the `url` in the IG resource [here](https://github.com/FHIR/sushi/blob/master/src/ig/IGExporter.ts#L133), and adding the `id` as well.

jafeltra:
> This _mostly_ works really well and I like it! The only issue I ran into is that if you populate _only_ the version in the configuration and then run GoFSH, you get this error:
> 
> > error Could not process StructureDefinition at Input_0: config is undefined
> 
> It works fine if you populate _only_ the Canonical. But _only_ the Version seems to be a problem.

I'm not entirely sure what the right way to resolve this is. This happens because I figured I'd only set the properties in the mini IG resource if they were set by the user, otherwise just let GoFSH generate the defaults. However, the problem is that GoFSH requires `url` to be present in the IG (see [here](https://github.com/FHIR/GoFSH/blob/master/src/processor/ConfigurationProcessor.ts#L16)). I feel like maybe there are two issues here:
1. The main issue for this PR is, what should FSH Online do? It doesn't seem like we can pass a version without passing a url, so should FSH Online not try to pass GoFSH the version if there is no canonical? This doesn't seem super intuitive since that would mean the two fields work slightly differently. Alternatively, FSH Online could pass "example.org" as the canonical if the version is set by the user. I don't know that I love either of those options, so I'm open to any others.
2. Separate from this PR, but GoFSH logs a warning if the IG resource is missing a url, but then doesn't return any configuration, so it throws an error trying to access properties off the `undefined` config, which is then caught and logged as the `Unexpected error...` you saw. There's probably a more graceful way for us to handle that.

src/components/FSHControls.js
```diff
@@ -139,7 +139,12 @@ export default function FSHControls(props) {
     props.resetLogMessages();
     props.onSUSHIClick(true, [''], true);
     const dependencyArr = sliceDependency(dependencies);
-    const config = { canonical, version, FSHOnly: true, fhirVersion: ['4.0.1'] };
+    const config = {
+      canonical: canonical ? canonical : 'http://example.org',
```
ngfreiter, line 143: Since it looks like the default value used by GoFSH is `sample.org` (see [here](https://github.com/FHIR/GoFSH/blob/e664582d8587109431fefff430ff877cedc5b079/src/extractor/ConfigurationExtractor.ts#L10)), I think we should just use `sample.org` here. It is a bit confusing if you put something like an Instance on the RHS, and then turn it into FSH, and see that now the configuration is populated with `sample.org`, when the default is `example.org`.

ngfreiter (commented):
In response to this [comment](https://github.com/FSHSchool/FSHOnline/pull/46#issuecomment-816667071), I think sending `example.org` (or `sample.org`) makes sense. We say that is the default value, so if I set the `version` but not `canonical`, I don't think it would be surprising that the default value would be used.

cmoesel:
We discussed this on Friday and determined that @jafeltra is going to make updates to GoFSH to use as much info as is passed along and infer the rest.  So if GoFSH is passed an IG JSON w/ `version` but not `url`, it will use the version from the IG JSON but infer the canonical from the other resources.  Once GoFSH works like that, then FSH Online will be able to properly support partial configurations.

Tagging @ngfreiter since this is related to his last comment.

src/components/FSHControls.js
cmoesel: I actually prefer `example.org`.  For one, it resolves to a more appropriate web site than `sample.org` (since `example.org` is actually owned and managed by IANA specifically to support use cases like this).

So I'd suggest that if we want to align better (and I think we do, so good call, @ngfreiter), then we should update GoFSH to use `example.org` instead.  Perhaps @jafeltra can do that as part of her other updates to GoFSH.

jafeltra:
Thanks for summarizing @cmoesel. I should have come back to update the PR after we talked. But now there's a GoFSH PR for this functionality here: https://github.com/FHIR/GoFSH/pull/108

I linked the GoFSH branch with this branch and setting only version works as expected.

src/components/FSHControls.js
jafeltra: Addressed here: 8873011

ngfreiter (commented):
This is looking good to me using that local version of GoFSH. The one thing that is a little weird is that if I run SUSHI with no configuration, the default values are used and the Configuration modal is untouched. But then if I run GoFSH even when my values perfectly match those default values, the modal is filled in. I'm not sure this even matters, I don't really have a strong opinion on whether those values should be filled in, but wanted to point it out in case you or @cmoesel did.

cmoesel:
@ngfreiter -- I think that's as designed.  I guess technically we _could_ check to see if the inferred values from GoFSH match the default values, but I'm not sure that's necessary or adds that much value.  I feel like when you run GoFSH on something, you're committing to whatever data is in the structures you're GoFSHing.  So... I'm fine w/ leaving it as it is.

ngfreiter (approved):
Given @cmoesel's comments, I approve.

cmoesel (approved):
I tried it w/ the version of GoFSH in your PR and it all works great! Nicely done!

PR #47, opened 2021-04-21
Update to new GoFSH Design

jafeltra:
This PR makes several changes to align with the designs from Noranda. There are general styling updates and theme updates to align with the general colors and theme that Noranda made (design [here](https://projects.invisionapp.com/share/HK10MES7J34D#/screens/448635151)). Along with that, there are a few functional updates:

- The JSON file definitions can be collapsed to the side and re-expanded. The JSON editor should grow and shrink as the list is moved. The list of definitions should scroll appropriately when the console is open or closed.
- The Configuration button is moved to the right hand side, but still works the same
- The Share button is moved to within the FSH editor, but it works the same
- The JSON editor displays the id of the definition that's being viewed in the header
- The delete buttons that were previously on each definition in the file list are removed. The delete is instead replaced with a single delete at the top of the JSON editor that deletes the current definition viewed in the editor.
- The FSH editor also has a delete button now
- When we eventually implement the Copy/Save actions, the action buttons should appear automatically.
- New FSH Online logo!
- I broke a few components out, including a general Delete Confirmation modal, the Share functionality, and a new component that renders the FSH editor now that it has more functionality, including the delete.
- The editor itself supports rendering a Drawer (right now, the JSON file list is the only drawer), but all functionality beyond opening and closing the drawer is controlled by the parent of the editor component (so in this case, JSON Output controls the Add Definition button, list of definitions, and switching definitions).
- Bonus: Added the GoFSH version next to the SUSHI version. This isn't based off designs so if you have strong (or mild) feelings about it's placement, let me know.

There are a lot of differences, but a good number of them are white space differences since everything shifts when I wrapped it in another component. Besides reviewing the code, try out the app and make sure the functionality is still as you'd expect. Let me know if you feel like anything from the designs is missing or if any of the implementation of the design feels off.


src/components/TopBar.js
```diff
@@ -1,62 +1,65 @@
...
+import { AppBar, Toolbar, Button, Typography, Box } from '@material-ui/core';
+import { StylesProvider } from '@material-ui/core/styles';
+import { ExitToApp } from '@material-ui/icons';
 import '../style/TopBarStyling.css';
+import logo from '../style/FSH-online-logo.png';
 
 const useStyles = makeStyles((theme) => ({
   root: {
-    background: '#2c4f85',
+    background: '#065171',
```
ngfreiter, line 11: This change makes it so that we no longer match the color of FSHSchool, which I thought was something we wanted for the top bar. I don't much mind, just wanted to make sure that was a conscious choice to move away from the matching FSHSchool.

ngfreiter (commented):
Should the favicon change to be the new logo as well? I vote probably yes, it looks a little odd to have the logo in the top left and then the non-matching logo as the icon.

I know the console wasn't necessarily so much a part of this, but while we are here, would we be able to update this: 
image

It just looks like everything on the console is all jammed together with no spaces. Could you add some spaces in there? And personally I'm not sure we need the word "success", I think the green check mark makes it clear.

src/components/TopBar.js
jafeltra: Yeah I consciously changed it to the color Noranda had in her design. The new color seems to be somewhere between the blue that's on the FSH School [landing page](https://fshschool.org/) (which was the old blue color here) and the blue of the top bar on the [Documentation pages](https://fshschool.org/docs/). I don't think I know enough about colors to say which way is better. Do you have any thoughts?

jafeltra:
> Should the favicon change to be the new logo as well?

I agree that it feels a little odd. I wasn't sure if I should balance keeping the same favicon as on FSH School or update it to match FSH Online. It is a little hard to see as a favicon, but it might just be because my browser is in dark mode.

Screen Shot 2021-04-22 at 2 10 45 PM


> It just looks like everything on the console is all jammed together with no spaces. Could you add some spaces in there? And personally I'm not sure we need the word "success", I think the green check mark makes it clear.

I'll add some padding around the icons and the words in the console! I'm a little hesitant to remove the words because there aren't any tooltips around the icons and I think we might want some explanation in case the symbols aren't clear, and hopefully the console will be updated relatively shortly after.

jafeltra:
Some updates!

- I added some space around the icons in the console. @ngfreiter let me know what you think. I didn't remove the words since I wasn't sure if it was worth adding in Tooltips and where to put the number associated with each, so I decided it isn't a regression from what's in the console now and will improve soon. Let me know if you agree.

image

- I talked to Noranda briefly, and she mentioned that @cmoesel thought the favicon should match the FSH School one since they're all based on the same site. I think I agree, but will also let @cmoesel weigh in on his own opinion. If we do want to have the favicon match the FSH Online logo, Noranda gave me a different one that will probably show up better than that screenshot I posted.

- For the color scheme, I think it depends on how integrated we want FSH Online to feel with FSH School. It does have its own logo and name separate from FSH School now, but if we want to keep it tied to FSH School through colors, I'll update it. I know we initially wanted that when it was a much simpler app, so I would also be curious about @cmoesel's feelings (since I've already pulled him in the second bullet). If no one has a strong feeling, I'm leaning toward the current blues.

ngfreiter:
> Some updates!

* Yay, like the spaces!
* Fineeee, but in that case my secondary comment is that they don't currently quite match:
image

FSH Online shorty is a little smooshed.

* I don't mind either way, I just thought we had wanted them to match.

jafeltra:
> FSH Online shorty is a little smooshed.

You got a problem with short Shorty??

I mean, fair - I never noticed that before.

jafeltra:
> FSH Online shorty is a little smooshed.

527032e unsmooshes him.

cmoesel:
Re: the favicon... I thought that you could only have a separate favicon if the site had its own subdomain (e.g., `fshonline.fshschool.org` could have its own favicon, but not `fshschool.org/FSHOnline`).  Under that assumption, I felt shorty was a more appropriate favicon for _the whole site_ than the fish-on-a-line icon would be.  BUT... if they can each have their own favicon, then I'm definitely open to having a FSHSchool-specific favicon.

jafeltra:
> Re: the favicon... I thought that you could only have a separate favicon if the site had its own subdomain

33 minutes ago I would have said I had no idea and you must be right that they're all the same, but since there were two different shorty favicons, I guess multiple are supported. I'll add in the FSH Online one so you all can see how it looks!

jafeltra:
> * I don't mind either way, I just thought we had wanted them to match.

3a6673c  Matched 'em! I think this probably does align better with the general FSH School feel.

And updated the favicon, which I don't think should overwrite the FSH School one since they've apparently been different this whole time. Let me know what you think!

src/components/TopBar.js
jafeltra: I matched them, and used the blue directly from the [documentation page](https://fshschool.org/docs/), so it's still different than it used to be, but I think it makes more sense since the top bars now match. Let me know if you prefer the original blue though.

ngfreiter:
> > * I don't mind either way, I just thought we had wanted them to match.
> 
> [3a6673c](https://github.com/FSHSchool/FSHOnline/commit/3a6673cfc170288549d9d3581cd6109d906a6f08) Matched 'em! I think this probably does align better with the general FSH School feel.
> 
> And updated the favicon, which I don't think should overwrite the FSH School one since they've apparently been different this whole time. Let me know what you think!

The favicon and the blue are looking good to me! I still have more to review so not approving yet, but thanks.

src/components/TopBar.js
cmoesel: +1 on top bars matching.  It looks great.

cmoesel (approved):
I don't feel qualified to give a full review of the code, but I did skim it and it all seemed at least reasonable to me!  Somebody who plays more often in React should probably take a closer look.

I built the static site and ran it and I love it.  I really love it.  It.  Is.  So.  Good.  I can't wait until everyone else gets to see it!

Gold star ⭐  and 1st place trophy 🏆 go to @jafeltra for this one!

julianxcarter (approved):
Both the code and the actual site look amazing to me, this is really well done @jafeltra! Nice work!

ngfreiter (approved):
Looking good 👍

PR #49, opened 2021-04-26
Description Update

jafeltra:
Updates the description that shows up in link previews (like in Slack). I know there are other fancy things you can do, like include images, but I figured I'd just start with this one place since it was pretty create-react-app-y.


cmoesel (approved):
Well water we wading for?  Let's get this merged!

PR #50, opened 2021-04-26
GoFSH in FSH Online

jafeltra:
This is it!

This PR contains the full GoFSH in FSH Online feature. All commits in this PR have been a part of PRs that were previously reviewed except for the final commit for merging master back into this branch (I'm not sure why it was needed, but it was). Even though individual pieces have been reviewed, it would probably be good to pull and run this one last time to make sure things still seem to be in order.

I think it still makes sense to squash and merge this PR into `master`, but since it's so large, I want to check with others before doing it.


cmoesel:
And it is PR #50. Was this your plan all along?  Masterfully executed.

cmoesel (approved):
It is time we allow the river of FSH to flow in both directions.*

_* This metaphor makes no sense, but luckily, the PR makes lots of sense._

PR #51, opened 2021-04-28
Only run SUSHI and GoFSH when not waiting for either

jafeltra:
This PR fixes a bug where clicking the Run SUSHI or Run GoFSH buttons while either was running would either crash the app or cause unexpected output. This updates it to only run either function when we are not waiting for any previous run. The button can still be clicked a bunch of times (@ngfreiter lookin @ u) but the output should be based on the first click.

The crash was originally coming from `JSON.parse`, so I wrapped that in a try/catch just to be safe, but the function should always get valid data now that other clicks are ignored.


ngfreiter (approved):
Now I can double click away!

PR #52, opened 2021-05-10
Design updates based on feedback

jafeltra:
This PR contains a few design updates based on feedback from Mark and Noranda, including:

- Making it clearer that there are several editor essentially one behind the other and tying the definition list more closely with the RHS JSON editor by updating the "active definition" color to match the color of the editor. (Note that this style is similar to OneNote, which Mark mentioned [here](https://mitrecorp.slack.com/archives/CSQV7N91A/p1619635053019800))
- Updating the scrollbar style. The main motivation for this is for Windows users. The Mac scrollbars disappear by default when you're not scrolling, but Windows's don't ,and they were the default white color, which also contributed to the disconnect between the editor and the definition list. This updates the style for both Mac and Windows (because I couldn't figure out a good way to only style a particular way on Windows). While the scrollbars don't disappear, they hopefully blend in more with the editor and don't visually break up the editor and definition list. 
- Changing the wording to "New JSON Editor" in order to not sound like we support file uploading and indicate that a new editor is added
- Clarifying that JSON should be input to the RHS editor by updating the placeholder text and more strongly indicate that only one JSON definition can be included in each editor
- Update the "Run" buttons to describe their function, not the tool they use behind the scenes
- Clarifying the Configuration button and configuration modal to better indicate that it is used for SUSHI and GoFSH configuration
- Sort new definitions (or any definition without a `resourceType`) into an "Unknown Type" group. All resources with a `resourceType` defined that is not a StructureDefinition, ValueSet, or CodeSystem are sorted to "Instances".
- Updating the overflow for definition ids in the definition list and on the editor header if it is too long to just use `...`.
- General styling updates including updates to button colors, hover colors, font weights, and padding

### DIFF
--- a/src/components/JSONOutput.js
+++ b/src/components/JSONOutput.js
@@ -245,7 +245,7 @@ export default function JSONOutput(props) {
     const id = defToDelete.id || 'Untitled';
     return (
       <DeleteConfirmationModal
-        title={'FHIR definition'}
+        title={'FHIR JSON'}
         item={`${type}/${id}`}
         isOpen={openDeleteConfirmation}
         handleCloseModal={handleCloseDeleteConfirmation}
@@ -255,10 +255,11 @@ export default function JSONOutput(props) {
   };
 
   const renderFileTreeView = () => {
-    const order = ['StructureDefinitions', 'ValueSets', 'CodeSystems', 'Instances'];
+    const order = ['StructureDefinitions', 'ValueSets', 'CodeSystems', 'Instances', 'Unknown Type'];
     const grouped = groupBy(fhirDefinitions, (val) => {
       if (['StructureDefinition', 'ValueSet', 'CodeSystem'].includes(val.resourceType)) return `${val.resourceType}s`;
-      return 'Instances';
+      if (val.resourceType != null) return 'Instances';
+      return 'Unknown Type';
     });
 
     return Object.keys(grouped)
@@ -312,7 +313,7 @@ export default function JSONOutput(props) {
     return (
       <>
         <Button className={classes.button} startIcon={<Add />} onClick={addDefinition}>
-          Create FHIR Definition
+          New JSON Editor
         </Button>
         <div className={classes.fileTreeContent}>{renderFileTreeView()}</div>
       </>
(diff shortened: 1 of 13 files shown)


src/components/JSONOutput.js
```diff
@@ -235,18 +255,19 @@ export default function JSONOutput(props) {
   };
 
   const renderFileTreeView = () => {
-    const order = ['StructureDefinitions', 'ValueSets', 'CodeSystems', 'Instances'];
+    const order = ['StructureDefinitions', 'ValueSets', 'CodeSystems', 'Instances', 'Unknown Type'];
     const grouped = groupBy(fhirDefinitions, (val) => {
       if (['StructureDefinition', 'ValueSet', 'CodeSystem'].includes(val.resourceType)) return `${val.resourceType}s`;
-      return 'Instances';
+      if (val.resourceType != null) return 'Instances';
+      return 'Unknown Type';
```
ngfreiter, line 262: Ok so I used my unique talents to make this crash here. I can't figure out how though. The error happened because `val` ended up undefined somehow. Can you see how that would happen? Maybe just to be safe you could return `Unkown Type` if `val` is undefined?

I'm sorry I can't give more info, but I'm really not sure how to replicate it regularly.

ngfreiter (commented):
Looks great! Pretty much approved, just one annoying thing.

src/components/JSONOutput.js
jafeltra: Hm yeah I'm struggling to see how/where that could be coming from. To be safe, I'll add in `val?.resourceType` in these two checks so that it will fall to the `Unknown Type` case if it's null. There are also a few other places to add `?.` to make it safe if a `fhirDefinition` is undefined.

jafeltra: This should be addressed in 23a750c. I'm still not sure what the original root cause was, but hopefully the addresses the symptoms.

ngfreiter (approved):
Looks good to me!

cmoesel (approved):
Looks and works great!

I quickly scanned the CSS, but mostly taking your work for it since it _looks_ good.

Issue #54, opened 2021-05-28
Bugs in text highlighting


markkramerus:
There are some anomalies in the text highlighting. See attached. 
![image](https://user-images.githubusercontent.com/492232/120003696-fcfdd480-bfa3-11eb-9654-77fe0078cc87.png)

markkramerus:
The "or" in discriminator should not be blue, the ".system" should be white, and perhaps the + in [+] should be blue.

markkramerus:
![image](https://user-images.githubusercontent.com/492232/120007149-94b0f200-bfa7-11eb-96b6-110ff305b90d.png)

The quoted codes are not highlighted correctly in this one.

cmoesel:
Thanks for reporting this.  We already have an open item in our backlog to fix highlighting bugs, so I've updated it to reference this issue.

jafeltra:
The "+" and quoted codes are properly highlighted now. They were fixed as part of #56.

The erroneous highlighting of "or" in discriminator and "system" as part of a path are symptoms of shortcomings in the way we implement the syntax highlighting. We have an open issue to evaluate a different approach to try to resolve these issues.

jafeltra:
I believe all the bugs in highlighting that are identified in this issue have been resolved and deployed to the site.

PR #56, opened 2021-06-04
Update FSH Syntax highlighting

jafeltra:
This PR updates the FSH syntax highlighting to more closely match what is in the VS Code extension. This adds a few new updates to the highlighting:

- Includes the `+` as a symbol (this is actually not supported in the VS Code extension, but I'll include it there as well)
- New keywords for Logical and Resource
- Highlights numbers
- Highlights quoted codes (`#"code with spaces"`)
- Highlight single quoted ucum units (`'lb'`)
- Highlight Reference and Canonical keyword. In the VS Code extension, we used more than just a regular expression to match the start and end parentheses without highlight the characters inside them. I'm not sure how to do that using the simple mode.

There are still three meaningful shortcomings:

- The simple mode we use for defining a language syntax in CodeMirror still does not support the lookbehind in regular expressions, so there are a few regular expressions that do not include that. This is the reason that `coding.system` will highlight system incorrectly.
- There are a few regular expressions that check that there is whitespace following the pattern. I haven't included this because CodeMirror doesn't seem to register a new line as white space, so I thought it would be better to have the value highlighting a few extra times rather than miss a few.
- The `\b` word boundary flag doesn't seem to be working as expected, which is what is leading to the `discriminator[0]` highlighting the `or`. I would have expected the `\b` on either side of the regular expression for the keywords to not match in that case.

These shortcomings might now be occurring often enough that we might want to consider expanding the language mode beyond the simple mode we are currently using. However, I thought it was better to get these fixes in first before embarking on that refactor.

This addresses some of the concerns in #54 but not all of them.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6777,12 +6777,16 @@
             "@babel/highlight": "^7.0.0"
           }
         },
+        "@babel/compat-data": {
+          "version": "7.14.0",
+          "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.14.0.tgz",
+          "integrity": "sha512-vu9V3uMM/1o5Hl5OekMUowo3FqXLJSw+s+66nt0fSWVWTtmosdzn45JHOB3cPtZoe6CTBDzvSw0RdOY85Q37+Q=="
+        },
         "@babel/core": {
           "version": "7.9.0",
           "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.9.0.tgz",
           "integrity": "sha512-kWc7L0fw1xwvI0zi8OKVBuxRVefwGOrKSQMvrQ3dW+bIIavBY3/NpXmpjMy7bQnLgwgzWQZ8TlM57YHpHNHz4w==",
           "requires": {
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
I've tried this and it works as advertised.  I agree that the behavior is odd (related to the `or` issue).  I tried a few other regular expression variations (like using `(^|\s)` in place of the first `\b` and `(\s|$)` in place of the last `\b` -- but those just exhibited the same problems.

Still, this PR leaves things in a better state than they were.  So I am approving.

cmoesel (changes requested):
Revoking my approval because I just realized that when I ran `npm install`, I ended up w/ a different `package-lock.json` than what is checked in here.  Can you check on that?

jafeltra:
@cmoesel I'm not sure why `package-lock.json` was out of date since this PR didn't change any packages, but I re-ran `npm install` and added the latest changes to `package-lock.json` here. I also merged in master, so this should be set again.

julianxcarter (approved):

cmoesel (approved):
Thanks, @jafeltra!

PR #57, opened 2021-06-04
Update to SUSHI 2.0 Beta

jafeltra:
~*This PR should be updated again when there is a new version of GoFSH to use. It is not ready to merge as it is now. I just wanted to put it up in case anyone wants to look at the SUSHI portion.*~ It's ready!

This PR updates to the latest version of SUSHI (and eventually GoFSH). In addition to updating the SUSHI version, I had to make minor changes to allow Logical Models and Resources be included in the console output and on the RHS output.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6774,11 +6774,6 @@
             "@babel/highlight": "^7.12.13"
           }
         },
-        "@babel/compat-data": {
-          "version": "7.14.0",
-          "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.14.0.tgz",
-          "integrity": "sha512-vu9V3uMM/1o5Hl5OekMUowo3FqXLJSw+s+66nt0fSWVWTtmosdzn45JHOB3cPtZoe6CTBDzvSw0RdOY85Q37+Q=="
-        },
         "@babel/core": {
           "version": "7.9.0",
           "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.9.0.tgz",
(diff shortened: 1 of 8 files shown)


cmoesel:
That's definitely a lot more changes than I expected!  Thanks!

What is the reason to wait for the new GoFSH release too?  Is it broken without that or are you just wanting to consolidate into a single release?

jafeltra:
They were pretty straightforward changes, which was nice!

I was thinking to wait for the GoFSH update so the two dependencies were in line with each other, but I don't think it really matters much since GoFSH will just be using the latest SUSHI as a dependency, but not include any new features. So I think that this actually should be fine without GoFSH!

cmoesel (commented):
I've tested this locally and it works well.  That said, I'm not giving my approval yet for two reasons:

1. There is a merge conflict that needs to be resolved.
2. It turns out maybe @jafeltra was right and we should wait for a new GoFSH.

Regarding 2 above, I tested this by entering a Logical Model FSH and Resource FSH.  It converted to FHIR JSON correctly.  But when I converted from JSON to FSH, it converted the logical models and resources to _Profiles_ (which are then horribly broken if you try to convert back to JSON).  I _expected_ it would convert them to `Instance`s of `StructureDefinition`.

Ideally, we would fix GoFSH to at least translate logical models and resources to instances rather than profiles before deploying this.

jafeltra:
> 1. There is a merge conflict that needs to be resolved.
This should be resolved now

> 2. It turns out maybe @jafeltra was right and we should wait for a new GoFSH.
I can look into this in GoFSH.

jafeltra:
Since GoFSH now has fixes in it that will help with the SUSHI 2.0 integration, I'm back to agreeing that we should wait for GoFSH 2.0 to merge this. I believe the branch should be set with SUSHI 2.0 now so that portion can be reviewed, but I'll update again when GoFSH 2.0 is available.

jafeltra:
This is ready to be re-reviewed, and should be set to be released, pending PR comments.

cmoesel:
I still get a different `package-lock.json` when I run `npm install` on this.  Do you @julianxcarter?

julianxcarter:
> I still get a different `package-lock.json` when I run `npm install` on this. Do you @julianxcarter?

I'm getting the same thing

jafeltra:
Ugh. I have no idea why this is happening. When I run `npm install`, I don't get any changes, and I had checked that the beta version of SUSHI was used in both the direct dependency and the GoFSH dependency, so I though it would be okay. Would one of you want to commit your `package-lock.json` to this branch? I'm not sure how to trigger an update for me.

jafeltra:
Actually I just swapped branches, installed, and then came back here to install, and I got some changes. I'll push them up, and hopefully that will resolve it. If not, maybe one of you could try.

cmoesel (approved):
SUCCESS!  And it all works well too.  Thanks, @jafeltra!

julianxcarter:
> Actually I just swapped branches, installed, and then came back here to install, and I got some changes. I'll push them up, and hopefully that will resolve it. If not, maybe one of you could try.
This did it for me! Thanks for fixing that.

julianxcarter (approved):

PR #59, opened 2021-06-07
Use hotfix version of SUSHI

jafeltra:
This PR reverts the revert commit so we get full support for LogicalModels and Resources back in the console output and the JSON output. It also uses the `2.0.0-beta.1-fshonline-hotfix` version of SUSHI in order to support `FlagRules` and `AddElementRules` in the browser.


cmoesel (approved):
Tested this via a static build locally and... it works!  Thank you for working late to get this fixed, @jafeltra!

PR #61, opened 2021-06-15
FSH Examples unit test

julianxcarter:
This PR adds a unit test checking that the Examples modal is properly populated from the `examplesConfig` prop passed into the `FSHControls` component. 

I wasn't able to get clicks on the actual example objects in the file tree to be recognized. This limited what I was able to actually test for, as I wasn't able to test that child examples are being rendered properly or that the actual example text is being fetched properly without being able to click objects in the file tree. I've included a second unit test in this PR (commented out) that shows the second test case that I think should be represented. If anyone has a workaround or knows of something that I'm doing wrong here, I'm all ears!

### DIFF
--- a/src/tests/components/FSHControls.test.js
+++ b/src/tests/components/FSHControls.test.js
@@ -406,51 +406,51 @@ it('should properly render the examples in the file tree', async () => {
   expect(manifestChild2).toBeInTheDocument();
 });
 
-// it('should populate editor when examples are collected', async () => {
-//   const updateTextValueSpy = jest.fn();
-//   document.body.createTextRange = () => {
-//     return {
-//       getBoundingClientRect: () => ({ right: 0 }),
-//       getClientRects: () => ({ left: 0 })
-//     };
-//   };
+it.skip('should populate editor when examples are collected', async () => {
+  const updateTextValueSpy = jest.fn();
+  document.body.createTextRange = () => {
+    return {
+      getBoundingClientRect: () => ({ right: 0 }),
+      getClientRects: () => ({ left: 0 })
+    };
+  };
 
-//   const manifestArr = [
-//     {
-//       id: 'manifestchild-1',
-//       name: 'manifestchild-1'
-//     },
-//     {
-//       id: 'manifestchild2',
-//       name: 'manifestchild-2'
-//     }
-//   ];
-//   const metadataObj = {
-//     'manifestchild-1': {
-//       name: 'manifestchild-1',
-//       description: 'First manifest object',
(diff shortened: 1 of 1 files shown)


ngfreiter (commented):
I tried and I also couldn't figure out how to make the tests work. I noticed that if the callback in the `TreeView` component was changed from `onNodeSelect` to `onClick`, it looked like it does trigger `fetchExamplesFSH`, but I don't want to have to change our code so that we can test it. I'd rather just figure out how the heck to test it. But I can't figure it out.

ngfreiter:
> I tried and I also couldn't figure out how to make the tests work. I noticed that if the callback in the `TreeView` component was changed from `onNodeSelect` to `onClick`, it looked like it does trigger `fetchExamplesFSH`, but I don't want to have to change our code so that we can test it. I'd rather just figure out how the heck to test it. But I can't figure it out.

I am not approving just yet though, because I'd like to see if maybe someone else can figure it out first, and if not, we can just take out the commented out piece and I'll approve.

julianxcarter:
> I tried and I also couldn't figure out how to make the tests work. I noticed that if the callback in the `TreeView` component was changed from `onNodeSelect` to `onClick`, it looked like it does trigger `fetchExamplesFSH`, but I don't want to have to change our code so that we can test it. I'd rather just figure out how the heck to test it. But I can't figure it out.

I'm not opposed to making minor changes to the actual code to make it a bit more testable. I did find that using `onClick` was problematic for the feature itself due to the lack of information that was available in the actual event. Somehow, the `fetchExampleFSH` function needs to take in the `id` of each node as an argument so that it can check the metadata lookup table for its GitHub url. I wasn't able to figure out how to pull this from a regular `click` event, but the custom `onNodeSelect` event that's specific to this component does provide the nodeID as an argument to its callback which allows the function to work. Switching to an `onClick` event could be a good solution if we can think of a way around that issue.

jafeltra (approved):
I couldn't figure out a way to properly test this either ☹️  I ran out of time to test out every thing I could think of, but it sounds like you both have tried some other options, so I feel fairly confident, and since I'll be out tomorrow, I don't want to block this for the rest of the week.

The one other thing you could do is uncomment the test and use a `it.skip` so we keep the skeleton of the test in case we ever want to come back to it in the future. I think I've done that with one or two other tests here. I don't want to block the merge just for that, so I'm approving as is.

ngfreiter (approved):
Looks good!

jafeltra (approved):
Someday maybe we will unskip the skips. 😴 💭   (<-those are clearly dreams)

PR #62, opened 2021-06-15
Resize Editors

### LINKED ISSUE #53: Support for re-sizable panels
Not sure if this is a desired feature, but it might be nice to be able to resize the FSH/FHIR panels in order to show more information on either end.

Someone using the site during a presentation had a tough time showing a long line while still being able to see the rest of the FSH content (i.e. without horizontal scrolling)

jafeltra:
This PR supports resizing the two editors. The center line can be dragged in either direction to resize either editor. Double clicking on the center line will recenter the editors back to their original placement.

This should work on touch for the most part (I was experimenting with just the Chrome Dev Tools), but the double click/double tap functionality doesn't work the same on touch. There look to be a few possible plugins that could help address this if we feel it is necessary.

I also could not get the "Convert" buttons to resize with the editor panes nicely. It is very straightforward to move the buttons along with the editor, but I'm not sure how to handle when the "Convert" buttons run into the "FSH Examples" and "Configuration" buttons. I really am not sure what the CSS would be to stop moving those buttons independently of the editor size at a certain location. I don't think we want that whole white bar to get taller to fit the buttons stacked on top of one another, so I didn't even explore that.

This fixes #53

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,7 @@
 import React, { createContext, useState, useEffect } from 'react';
 import { inflateSync } from 'browserify-zlib';
+import { debounce } from 'lodash';
+import clsx from 'clsx';
 import { createMuiTheme, makeStyles } from '@material-ui/core/styles';
 import { Grid, ThemeProvider } from '@material-ui/core';
 import { expandLink } from './utils/BitlyWorker';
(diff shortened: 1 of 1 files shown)


ngfreiter (approved):
This looks good to me, I think it's fine to have the static buttons at the top.

julianxcarter (approved):
This looks great! The double click to reset was a very nice touch, and I also agree that the static buttons aren't really a concern.

PR #63, opened 2021-06-21
Realign Convert Buttons

jafeltra:
Because the dividing space between the two editors got wider, the space between the two convert buttons no longer aligns. It really bugs me and was easy to fix, so I made a PR for it. I'll add screenshots of the differences, but you can also see the difference by looking at https://fshschool.org/FSHOnline/#/ and this branch.

On the deployed FSH Online site now, the space between the buttons does not align with the space between the editors:

image

On this branch, it does align:

image

I think they might still be a hair off, but this feels much better to me.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6774,12 +6774,19 @@
             "@babel/highlight": "^7.12.13"
           }
         },
+        "@babel/compat-data": {
+          "version": "7.14.0",
+          "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.14.0.tgz",
+          "integrity": "sha512-vu9V3uMM/1o5Hl5OekMUowo3FqXLJSw+s+66nt0fSWVWTtmosdzn45JHOB3cPtZoe6CTBDzvSw0RdOY85Q37+Q=="
+        },
         "@babel/core": {
           "version": "7.14.2",
           "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.14.2.tgz",
           "integrity": "sha512-OgC1mON+l4U4B4wiohJlQNUU3H73mpTyYY3j/c8U9dr9UagGGSm+WFpzjy/YLdoyjiG++c1kIDgxCo/mLwQJeQ==",
           "requires": {
+            "@babel/code-frame": "^7.12.13",
             "@babel/generator": "^7.14.2",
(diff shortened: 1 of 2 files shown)


jafeltra:
Plz ignore the massive blunder above

cmoesel:
I guess I have no choice but to review this now.  Julia says I _must_.

cmoesel (dismissed):
This does look good and I approve.

Would it drive you crazy if I posted this super zoomed version that shows the actual "hair" of difference? 
image

;-)

I tried to see if it was easily fixable.  It is not.  But I think it is definitely good enough for normal people who are not maniacs like me.

Also, it can't be related to this PR since you did not change `package.json`, but when I run `npm install`, I am getting a different `package-lock.json` than what is checked in. :-/  I think it's harmless, but thought I'd mention it.

jafeltra:
> Would it drive you crazy if I posted this super zoomed version that shows the actual "hair" of difference?

Yes.

But also, I did do that crazy zoomed in thing which is why I knew the hair difference was there, and I was hoping no one else would investigate the hair difference. I also tried briefly to fix it, but was unsuccessful. I'll try one more thing.

> when I run npm install, I am getting a different package-lock.json than what is checked in.

Ugh I'm not sure if this has to do with me or not, but this is not the first time it's happened in this repo and I'm not sure why it keeps happening. I'll at least run `npm install` and check in the changes.

jafeltra:
Thanks to @ngfreiter the difference is gone! 😅 It actually looks even better than it did in the pre-resizable editor world.

I also added the changes from `npm install` but I keep having issues with my `npm install` not getting all the changes, so if you still see differences, let me know.

cmoesel (approved):
Amazing.  The @ngfreiter... he's not so bad.

image

Fractional pixels.  Who knew?  What even is a pixel anymore?

ngfreiter (approved):
Looks good and gave me a chance to update my resume
Screen Shot 2021-06-21 at 4 36 11 PM

jafeltra:
I'm shocked that fractional pixels seems to be the one thing that is supported uniformly across all browsers.

> Yes, you can specify fractional pixels. As this has been part of CSS since the very first version, it should be well supported by any browser that supports CSS at all.

However, I'll let the internet just do it's thing this time around.

PR #64, opened 2021-07-13
Update JSON file list to organize by FSH entity

jafeltra:
This PR updates the organization of the JSON files we display on the far right side of the page. Instead of grouping definitions by the FHIR `resourceType`, they are now grouped by the FSH entity type that they came from or will become when running GoFSH. This reflects the default output style of GoFSH once we release a new version. I don't think that the GoFSH release should hold up the merge of this PR since it is still a valid way to organize the files.

The one area I'd definitely like feedback on is if the criteria I'm using for differentiating between Profiles, Extensions, Logical Models, and Resources is sufficient. I used the criteria from GoFSH [here](https://github.com/FHIR/GoFSH/blob/master/src/processor/LakeOfFHIR.ts#L92-L112) (which I believe @mint-thompson added recently). It seems to be working as expected, but I could be missing something.

I also updated the size of the box that SUSHI outputs to match the narrower box in SUSHI.

### DIFF
--- a/src/components/JSONOutput.js
+++ b/src/components/JSONOutput.js
@@ -78,6 +78,41 @@ const useStyles = makeStyles((theme) => ({
   }
 }));
 
+const checkFshType = (def) => {
+  // The criteria in this function is based on isSDForInstance in GoFSH (LakeOfFHIR.ts)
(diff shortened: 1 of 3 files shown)


ngfreiter (approved):
Approved!

julianxcarter (approved):
Looks good!

PR #65, opened 2021-07-19
Support sharing configuration in links

ngfreiter:
This addresses [CIMPL-711](https://standardhealthrecord.atlassian.net/browse/CIMPL-711) by adding support for sharing configuration information when sharing links. If there is information entered into the config modal, and you share a set of FSH, when you visit the generated link, the configuration information should be prepopulated with what was there when you generated the link.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -196,10 +196,16 @@ export default function App(props) {
   const [exampleFilePaths, setExampleFilePaths] = useState({});
   const [leftWidth, setLeftWidth] = useState(41.666667); // Initial width based off grid item xs={5} size to align with FSHControls
   const [isDragging, setIsDragging] = useState(false);
+  const [config, setConfig] = useState();
 
   useEffect(() => {
     async function waitForFSH() {
-      setInitialText(await decodeFSH(urlParam.params));
+      const text = await decodeFSH(urlParam.params);
+      const splitIndex = text.indexOf('\n');
+      const config = text.slice(0, splitIndex);
+      const fshContent = text.slice(splitIndex + 1);
+      setConfig(JSON.parse(config));
+      setInitialText(fshContent);
     }
     async function fetchExamples() {
       setExampleConfig(await getManifestFromGit());
@@ -238,6 +244,10 @@ export default function App(props) {
 
   function updateInputFHIRTextValue(text) {
     setInputFHIRText(text);
+  }
+
+  function handleConfigChange(config) {
+    setConfig(config);
   }
 
   function handleResetWidth() {
@@ -293,6 +303,8 @@ export default function App(props) {
           <FSHControls
             onSUSHIClick={handleSUSHIControls}
             onGoFSHClick={handleGoFSHControls}
+            onConfigChange={handleConfigChange}
+            config={config}
             fshText={inputFSHText}
             gofshText={inputFHIRText}
             resetLogMessages={resetLogMessages}
@@ -322,6 +334,7 @@ export default function App(props) {
                   updateTextValue={updateInputFSHTextValue}
                   isWaiting={isWaitingForFSHOutput}
                   setInitialText={setInitialText}
+                  config={config}
                 />
               </Grid>
               <Grid
--- a/src/App.js
+++ b/src/App.js
@@ -196,7 +196,8 @@ export default function App(props) {
   const [exampleFilePaths, setExampleFilePaths] = useState({});
   const [leftWidth, setLeftWidth] = useState(41.666667); // Initial width based off grid item xs={5} size to align with FSHControls
   const [isDragging, setIsDragging] = useState(false);
-  const [config, setConfig] = useState();
+  const [configToShare, setConfigToShare] = useState({});
+  const [sharedConfig, setSharedConfig] = useState({});
 
   useEffect(() => {
     async function waitForFSH() {
@@ -204,7 +205,7 @@ export default function App(props) {
       const splitIndex = text.indexOf('\n');
       const config = text.slice(0, splitIndex);
       const fshContent = text.slice(splitIndex + 1);
-      setConfig(JSON.parse(config));
+      setSharedConfig(config ? JSON.parse(config) : {});
       setInitialText(fshContent);
     }
     async function fetchExamples() {
@@ -247,7 +248,7 @@ export default function App(props) {
   }
 
   function handleConfigChange(config) {
-    setConfig(config);
+    setConfigToShare(config);
   }
 
   function handleResetWidth() {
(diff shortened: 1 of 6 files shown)


src/components/FSHControls.js
```diff
@@ -98,6 +98,16 @@ export default function FSHControls(props) {
   const [currentExample, setCurrentExample] = useState('');
   const [currentExampleName, setCurrentExampleName] = useState('');
 
+  if (!canonical && props.config?.canonical) {
+    setCanonical(props.config?.canonical);
```
jafeltra, line 102: Is the `?.` necessary here if it has made it into the if statement?

jafeltra (commented):
This looks good! I just had a couple small comments:

- When I load a share link that didn't contain any configuration, I get this error in the console: 


```
Uncaught (in promise) SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at waitForFSH (App.js:208)
```

src/components/FSHControls.js
ngfreiter: Fixed in https://github.com/FSHSchool/FSHOnline/pull/65/commits/56b12ac7d49d2f74b37c0ef479fb0d76f83b18e8

src/App.js
```diff
@@ -196,10 +196,16 @@ export default function App(props) {
   const [exampleFilePaths, setExampleFilePaths] = useState({});
   const [leftWidth, setLeftWidth] = useState(41.666667); // Initial width based off grid item xs={5} size to align with FSHControls
   const [isDragging, setIsDragging] = useState(false);
+  const [config, setConfig] = useState();
```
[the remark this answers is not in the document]
ngfreiter: Yeah, sorta done in https://github.com/FSHSchool/FSHOnline/pull/65/commits/56b12ac7d49d2f74b37c0ef479fb0d76f83b18e8, but I realized that there really are two configs that need to exist in the `App.js` file. One is the config that is shared, and one is the config that is going to be shared. If these two are not separated, then after visiting a page with a shared config, you can't blank out the config without it being reset back to the original config.

src/components/ShareLink.js
```diff
@@ -41,7 +41,7 @@ export default function ShareLink(props) {
   };
 
   const handleOpenShare = async () => {
-    const encoded = deflateSync(props.shareText).toString('base64');
+    const encoded = deflateSync(JSON.stringify(props.config || {}) + '\n' + props.shareText).toString('base64');
```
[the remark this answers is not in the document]
ngfreiter: Done in https://github.com/FSHSchool/FSHOnline/pull/65/commits/56b12ac7d49d2f74b37c0ef479fb0d76f83b18e8.

ngfreiter:
> This looks good! I just had a couple small comments:
> 
> * When I load a share link that didn't contain any configuration, I get this error in the console:
> 
> ```
> Uncaught (in promise) SyntaxError: Unexpected end of JSON input
>     at JSON.parse (<anonymous>)
>     at waitForFSH (App.js:208)
> ```

Error should be gone in https://github.com/FSHSchool/FSHOnline/pull/65/commits/56b12ac7d49d2f74b37c0ef479fb0d76f83b18e8

src/App.js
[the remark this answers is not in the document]
jafeltra: Oh whoa that's trippy. Good catch!

jafeltra (approved):
This looks good to me! Nifty feature!

ngfreiter:
> Uh oh! I'm sorry for the 🎢 but I think I take back my approval. If I try to use a share link that was generated before this feature was added, I get an error trying to parse the FSH and config info, which I think is because there's no `\n` to properly split on. I'm not sure what the best way to address this is (I haven't thought it through yet), but I don't think we want those old links to break when this gets merged.

Ok I think https://github.com/FSHSchool/FSHOnline/pull/65/commits/bce170e9af05e42858fae5fc88450cb2bf56d728 should fix the issue with sharing legacy links, since now we only assume the first line represents config if we can parse it as JSON, and if it contains the expected properties.

cmoesel:
My only concern w/ this feature is that it reduces the amount of FSH we can share (since the overall content length is finite).  To that end, do you think it would be OK to shorten the encoded config properties so they are `{ c: '', v: '', d: '' }`?  Or is that silly?  Also, what happens if the user has not modified the config (default values)?  Is it encoding those default values or is that what `{}` is representing?

ngfreiter:
> My only concern w/ this feature is that it reduces the amount of FSH we can share (since the overall content length is finite). To that end, do you think it would be OK to shorten the encoded config properties so they are `{ c: '', v: '', d: '' }`? Or is that silly? Also, what happens if the user has not modified the config (default values)? Is it encoding those default values or is that what `{}` is representing?

That's a fair ask, doesn't hurt to keep the shareable size as big as possible. If the config is not modified, we send something like this:
```
{ canonical: '', version: '', dependencies: '' }
```

ngfreiter:
https://github.com/FSHSchool/FSHOnline/pull/65/commits/3a8ff764d2a097490f28b55ec534bb5ef9f5321a makes the config smaller, and fixes an issue with me doing something bad which would cause the config values to be impossible to blank out if they were set from a shared link.

src/components/ShareLink.js
```diff
@@ -41,7 +41,8 @@ export default function ShareLink(props) {
   };
 
   const handleOpenShare = async () => {
-    const encoded = deflateSync(JSON.stringify(props.config) + '\n' + props.shareText).toString('base64');
+    const itsyBitsyConfig = { c: props.config?.canonical, v: props.config?.version, d: props.config?.dependencies };
```
jafeltra, line 44: One question/idea I had was if we could make this even itsy bitsier. If there is no canonical or version or dependencies to share, does it make sense to not include any config + `\n` in the string that we compress?

jafeltra (commented):
This looks really good to me! The old links still work and I can load and replace and clear configuration after sharing. I just had one other question that I asked below.

src/components/ShareLink.js
ngfreiter: Yeah I think that is fair, why not save the space. And it should work fine since we've already supported that case for legacy links. I'll make the change.

ngfreiter: Should be good in https://github.com/FSHSchool/FSHOnline/pull/65/commits/f1021072b93b47287cae09d785c7739f17d096ef.

jafeltra (approved):
This looks great! The old links work, and I think this is the smallest amount of information we can add to the URL given the current approach.

cmoesel:
I don't know if this is related to your PR or something else funky going on w/ my environment.  When I tested this, I updated the config to have dependency: `hl7.fhir.us.core#4.0.0` (but did not set canonical or version).  I added the following FSH:
```
Profile: MyPatient
Parent: USCorePatientProfile
* name.given 1..1
```

When I clicked Convert to JSON the first time, it seemed like it didn't take the dependency (it couldn't resolve USCorePatientProfile), so I tried it again and then got:
image

cmoesel (approved):
OK.  We determined that the bug above is not related to this PR and @jafeltra (in all her brilliance) will fix it in a different PR.  So... approved!

PR #68, opened 2021-08-16
Update to use GoFSH 1.3.0

ngfreiter:
This updates FSHOnline to use GoFSH 1.3.0. Some of the functions changed slightly and needed to be updated, and the boxes at the end of the output also needed to be edited to match what GoFSH now logs.

### DIFF
--- a/src/utils/FSHHelpers.js
+++ b/src/utils/FSHHelpers.js
@@ -12,7 +12,6 @@ const stats = utils.stats;
 const getRandomPun = utils.getRandomPun;
 const Type = utils.Type;
 const FHIRDefinitions = fhirdefs.FHIRDefinitions;
-const { getRandomSeaCreatures, getRandomSeaCreaturesStat } = gofshUtils;
 
 /**
  * Run GoFSH
@@ -40,7 +39,7 @@ export async function runGoFSH(input, options) {
       return;
     }
     if (gofshUtils.isProcessableContent(resource, location)) {
-      docs.push(new processor.WildFHIR(resource, location));
+      docs.push(new processor.WildFHIR({ content: resource }, location));
     }
   });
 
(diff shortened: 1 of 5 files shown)


src/utils/FSHHelpers.js
```diff
@@ -12,7 +12,6 @@ const stats = utils.stats;
 const getRandomPun = utils.getRandomPun;
 const Type = utils.Type;
 const FHIRDefinitions = fhirdefs.FHIRDefinitions;
-const { getRandomSeaCreatures, getRandomSeaCreaturesStat } = gofshUtils;
```
jafeltra, line 15: Bye sea creatures. 👋

jafeltra (approved):
Looks good to me! It's nice to have the round trip back!

jafeltra:
Gah wait I thought of something else. Since we now support profiles of logical models, we might want to update the sorting we do as a part of this PR.

ngfreiter:
> Gah wait I thought of something else. Since we now support profiles of logical models, we might want to update the sorting we do as a part of this PR.
> 
> **Update**: I actually think this should be very easy, so I vote we add it into this PR. The function `checkFshType` [here](https://github.com/FSHSchool/FSHOnline/blob/master/src/components/JSONOutput.js#L81) just needs to be udpated so that we include `'logical'` as an allowed `kind` for profiles. So I think we can just update [line 100](https://github.com/FSHSchool/FSHOnline/blob/master/src/components/JSONOutput.js#L100) to be
> 
> ```
> } else if (['resource', 'complex-type', 'primitive-type', 'logical'].includes(def.kind)) {
> ```

Good catch! Done in https://github.com/FSHSchool/FSHOnline/pull/68/commits/b8582a50bf105e3545ea0601014e364622903f5d.

jafeltra (approved):
This looks good! Thanks for making that change!

julianxcarter (approved):

PR #70, opened 2021-09-01
Support sharing FSH content via Gist links

ngfreiter:
## Description

This adds support for loading content into FSH Online from a Gist. If a URL is of the form:
```
<base>/FSHOnline/#/gist/<gistId>
```
then FSHOnline will load the content of the Gist at `<gistid>` into the text editor. For example:
```
<base>/FSHOnline/#/gist/a580d9af4e81db73a26d9bbc75f428e7
```
will load this [gist](https://gist.github.com/ngfreiter/a580d9af4e81db73a26d9bbc75f428e7) into the editor. If there are multiple files in the gist they will just be appended together.

There is also now an option to create a link from a Gist URL in the "share" options. I think we had discussed the idea of copying the content of the editor to the clipboard automatically when going to create a Gist, but I couldn't really figure out a way to designing that in a way that made sense to me, outside of just adding a "copy" option alongside "share" and "trash".

## Testing
Test that 
* You can still generate a direct link
* Generating too long of a direct link is still an error
* You can load FSH from a Gist
* You can load FSH from a multi-file Gist
* You can load FSH from a Gist that is long enough to be "truncated" (you'll be able to tell when making the Gist since Github will tell you if the content you are viewing is "truncated")
* You can create a share URL from a Gist URL
* You can copy the share URL to your clipboard

I wanted to be able to test the stuff added to `App.js`, but I had a lot of trouble figuring out how to mock the `fetch` method, and ultimately couldn't do it. If anyone is able to figure that out, let me know and I can get those tests in there.


src/components/ShareLink.js
```diff
@@ -60,56 +85,89 @@ export default function ShareLink(props) {
...
+            <DialogContent>
+              Generate a link directly from FSH Online to share your FSH with others. This option is best for small
+              snippets of FSH.
+            </DialogContent>
+            <DialogActions>
+              <Button onClick={handleGenerateLink} color="primary">
+                Generate Direct Link
+              </Button>
+            </DialogActions>
+          </Grid>
```
jafeltra, line 120: I'd definitely like your opinion on how difficult this is and if it is worth doing as part of this PR, but I think it might be nice for users to know right away if their FSH is too long to be shared with a direct link. It might be nice to have this section conditionally render what it currently says if the FSH is short enough or render a message that their FSH is too long and should be shared with a Gist instead and then we can disable the Generate Link button here as well.

src/App.js
```diff
@@ -180,6 +180,20 @@ export async function decodeFSH(encodedFSH) {
...
+async function getGistContent(gistId) {
+  const gistContent = await fetch(`https://api.github.com/gists/${gistId}`).then((res) => res.json());
+  let fshContent = '';
+  for (let f of Object.values(gistContent?.files || {})) {
+    if (f.truncated) {
+      fshContent += await fetch(f.raw_url).then((res) => res.text());
+    } else {
+      fshContent += f.content;
+    }
+    fshContent += '\n';
```
jafeltra, line 192: Should there be two new lines between separate files? If I have a FSH entity in each file, I think we want them to be separated by an extra new line, right?

jafeltra (commented):
I tried out all the test cases you described and it works great! I had a few small comments below. I'm also going to try out the mock for fetch but haven't had a chance to look yet.

src/components/ShareLink.js
ngfreiter: I think this makes sense, it means we will generate more links, but if we weren't bumping into the API limits before, I don't think we will now, and if we do we can always change it up, so sounds good!

src/App.js
```diff
@@ -201,25 +215,30 @@ export default function App(props) {
...
+          // set the parsedConfig, and set fshContent to all of the text
         }
-      } catch (e) {
-        // If parse fails, it is likely decoding a legacy link in which all content is FSH, so just don't
-        // set the parsedConfig, and set fshContent to all of the text
+        setSharedConfig(parsedConfig || {});
+        setInitialText(fshContent);
+      } else if (/\/gist/.test(urlParam.path)) {
+        const fshContent = await getGistContent(urlParam.params.id);
+        setInitialText(fshContent);
```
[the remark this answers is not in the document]
ngfreiter: Yeah I think this is a good idea, but I would vote separate PR, since I think we need to do some more thinking about how it would work.

cmoesel:
Ha.  `get-the-gist`.  That's funny.

cmoesel (changes requested):
This is definitely cool.  The functionality is quite useful indeed.  It also was fun to put random numbers in and occasionally get something interesting.  For example, this not so random number: http://localhost:3000/FSHOnline/#/gist/1

I was actually surprised that username is not part of the gist identifier -- since the gist URLs all seem to have the username in them. But... it seems to work, so I'm assuming this is legit.

The one thing I'd like us to consider is changing the modal itself.  I can't pinpoint exactly what the problem is, but I find it visually confusing.  It could perhaps be improved by increasing the size of the modal (so the text doesn't wrap so much).  It also wasn't clear what the text-box looking thing at the bottom of the modal was when I first opened it. We may want to either put in a label/placeholder text (e.g., "Shareable link will be generated here") or hide the input box until you generate the link.  Last, I think the standard link sharing is the easiest and preferred way -- so I feel like we should give that the primary focus, and maybe de-emphasize the gist approach (but still definitely make it available)... If we don't mind delaying this PR, I'd suggest perhaps throwing the modal at Noranda for her thoughts.

src/App.js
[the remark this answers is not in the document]
cmoesel: The first thing that comes to mind is using some kind of special annotation in comments at the top of the gist.  But... for multi-file gists I guess you would need to loop through all the files looking for it (and hope it doesn't appear more than once)?

Another approach would be to require a multifile gist where one of the files is a specific config file w/ the info needed.

And the last idea that would _partially_ work is to attempt to _detect_ dependencies.  Basically, we would keep a map of canonicals for the most popular IGs, then grep the FSH for use of those canonicals -- and derive dependencies for any that are found.  The tricky bit is determining what version(s) for the IG we should use.  We could default to the latest version, but then we need to keep up to date regarding latest versions of those IGs (unless there is an API we can use).  We could also prompt the user that we detected these dependency IGs and have them choose or confirm the versions...

src/components/ShareLink.js
cmoesel: If bit.ly publishes their max supported URL size, we might be able to detect this before attempting to create the link.  That said, I'm not sure they publish the max size (otherwise we might have already used it by now).  Still... might be worth looking into again.

ngfreiter:
@cmoesel and @jafeltra, https://github.com/FSHSchool/FSHOnline/commit/0780e82215ff0c75c62077db37ffa38348f66519 should address those issues @cmoesel found with the design. Let me know if you have additional thoughts.

src/components/ShareLink.js
```diff
@@ -60,56 +85,89 @@ export default function ShareLink(props) {
...
+              <TextField id="gistURLText" label="Gist URL" fullWidth onChange={updateGistURL} />
+            </DialogContent>
+            <DialogActions style={{ bottom: 0, right: 0 }}>
+              <Button
+                disabled={gistURL === ''}
+                onClick={handleGenerateLinkFromGist}
+                color="primary"
+                style={{ whiteSpace: 'nowrap' }}
+              >
+                Generate Link from Gist
```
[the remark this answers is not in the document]
ngfreiter: No longer really relevant, since the two options aren't side by side.

src/components/ShareLink.js
```diff
@@ -60,56 +85,89 @@ export default function ShareLink(props) {
...
+          </Grid>
+          <Grid item xs={12}>
+            <DialogContent>
+              <TextareaAutosize
+                id="link"
+                disabled
+                label="Your Link"
+                onChange={updateLink}
+                className={classes.textArea}
+                value={link}
```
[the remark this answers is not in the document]
ngfreiter: Should be fixed in https://github.com/FSHSchool/FSHOnline/commit/0780e82215ff0c75c62077db37ffa38348f66519.

src/components/ShareLink.js
```diff
@@ -60,56 +85,89 @@ export default function ShareLink(props) {
...
+            </DialogActions>
+          </Grid>
+        </Grid>
       </Dialog>
       <Dialog open={openShareError} onClose={handleCloseShareError} aria-labelledby="alert-dialog-title" maxWidth="lg">
         <DialogTitle id="alert-dialog-title">Share Error</DialogTitle>
         <DialogContent>
-          <DialogContentText>There was a problem sharing your FSH. Your FSH file may be too long.</DialogContentText>
+          <DialogContentText>
+            There was a problem sharing your FSH. Your FSH file may be too long. Consider sharing your FSH with a Gist.
```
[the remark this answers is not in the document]
ngfreiter: Again no longer an issue since things have changed

src/components/ShareLink.js
ngfreiter: Addressed in https://github.com/FSHSchool/FSHOnline/commit/0780e82215ff0c75c62077db37ffa38348f66519.

src/App.js
ngfreiter: Forgot to do this in https://github.com/FSHSchool/FSHOnline/commit/0780e82215ff0c75c62077db37ffa38348f66519.

ngfreiter: But remembered to do it in https://github.com/FSHSchool/FSHOnline/pull/70/commits/f148a0fd66f27a5025d7a183519a2ed6c6c02fa9.

cmoesel:
I like this.  The user sees the simple approach foremost and only needs to get into the gisty details if they choose to or if their FSH is too long.  Well done.

I only have one suggestion for a possible improvement.  If the user puts in FSH that is too long, they get this dialog:
image

Now... if you think about the workflow, in order to actually do what we suggest, the user will need to first _close the dialog_, then copy their text from the FSH pane, then click the link button again (at which point we'll try to create a normal link again), and then finally create their gist.  I'd suggest we add a button or link in the dialog to copy the FSH to the clipboard right there so they don't have to close out of the dialog.  What do you think of that idea, @ngfreiter and @jafeltra?

ngfreiter:
> I like this. The user sees the simple approach foremost and only needs to get into the gisty details if they choose to or if their FSH is too long. Well done.
> 
> I only have one suggestion for a possible improvement. If the user puts in FSH that is too long, they get this dialog:
> image
> 
> Now... if you think about the workflow, in order to actually do what we suggest, the user will need to first _close the dialog_, then copy their text from the FSH pane, then click the link button again (at which point we'll try to create a normal link again), and then finally create their gist. I'd suggest we add a button or link in the dialog to copy the FSH to the clipboard right there so they don't have to close out of the dialog. What do you think of that idea, @ngfreiter and @jafeltra?

Yeah I think that makes sense, added a button in https://github.com/FSHSchool/FSHOnline/pull/70/commits/9420e3fe34af43ef820fd663729f28de9f7c280f.

cmoesel (dismissed):
Lookin' good!

ngfreiter:
dca5346 changes the modal per some of the offline discussions we had. It's a combination of suggestions from both @cmoesel and @jafeltra, so both let me know what you think.

cmoesel:
I like the new approach.  It seems very consistent and is easy to use!

I noticed that my gist share link wrapped in the dialog:
image

If gist IDs are fairly predictable sizes (I'm guess they are?) then do you think we might want to increase the modal width so we can display them without wrapping?  Or do you think that would make the modal too wide?

jafeltra:
This looks great to me! I really like being able to see the link we generate and I really like the new copy button, so thank you for making those changes! I didn't notice anything else besides the size of the modal already discussed above!

ngfreiter:
@jafeltra and @cmoesel, updated the `maxWidth` of the modal in https://github.com/FSHSchool/FSHOnline/pull/70/commits/37f4a3882f5bbc85f167f00fab7cb7af82cc5962.

src/App.js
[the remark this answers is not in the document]
jafeltra: I created a separate task for this: https://standardhealthrecord.atlassian.net/browse/CIMPL-833

PR #71, opened 2021-09-01
Syntax Highlighting Updates

jafeltra:
I made some minor updates to the syntax highlighting for FSH that our CodeMirror editor uses.

I looked into using the full language mode (rather than the simple mode), but it was requiring parsing a lot of the text stream and I didn't feel confident that I would be able to catch and handle every possible case. Instead, I found some ways to improve our existing regular expressions, and I think I've covered all of the shortcomings that we know of. I also found a way to highlight the opening and closing parentheses around `Reference` and `Canonical`.

The FSH here is pretty meaningless, but [here](https://fshschool.org/FSHOnline/#/share/3yFELYy) is a link with FSH to test out some of the edge cases. You can view this on fshschool and using this branch and see the differences in the highlighting.

This should address the remaining issues listed in #54.


julianxcarter (approved):
This looks good! I can confirm that all of the problems discussed in Issue #54 have been addressed between this PR and the first swing that you took in June. It's awesome that you were able to make it happen using the simple language mode too. Great work @jafeltra!

cmoesel (approved):
Wow. So much better!  Good work figuring that out!

PR #73, opened 2021-09-21
Update dependencies

jafeltra:
In this PR, I updated dependencies using `npm audit fix` and updated `react-scripts` and the react testing library manually. Updating `react-scripts` and `@testing-library/react` resolved various vulnerabilities reported. I had to update how we mock out the `getBoundingClientRect` function and I added it to the start up script for the tests so it just run before all the tests are run. I also replaced the `wait` function with `waitFor` when updating the react testing library.

If you run `npm install`, there should be changes in the package-lock.json file, and a few additional vulnerabilities. When you run `npm audit fix` after that, the changes should go away. I'm not sure why npm audit fix works this way, but it is updating the dependencies inside SUSHI and GoFSH (which are already updated in the repos, but not released yet). However, these updates aren't made when just running `npm install`. Instead of running an `npm install`, you can run an `npm ci` to use what it is specified in the package-lock.json, or you can follow the `npm install` and `npm audit fix` pattern and ensure you have no changes.


ngfreiter (dismissed):

julianxcarter (approved):
Looks good to me!

jafeltra:
@ngfreiter and @julianxcarter - I resolved the conflicts that were on this branch, which dismissed your reviews. I only updated new tests with the same changes I made in existing tests. If one of you re-approves, I think this should be good to go.

ngfreiter (approved):

PR #74, opened 2021-09-22
Problems tab in the Console

julianxcarter:
This PR splits the errors and warnings out of the main console and into a separate "Problems" column, as seen in [Noranda's initial design](https://projects.invisionapp.com/share/HK10MES7J34D#/screens/448635151). I thought this made the `Success`, `Warnings`, and `Errors` indicators obsolete, but if we think they should still be included in some fashion then they can be added back.


src/App.js
```diff
@@ -230,11 +229,8 @@ export default function App(props) {
   }, [urlParam]);
 
   function resetLogMessages() {
-    consoleMessages = [];
-    errorString = '';
-    warningString = '';
-    errorCount = 0;
-    warningCount = 0;
+    infoMessages = [defaultInfoMessage];
+    problemMessages = [defaultProblemMessage];
```
ngfreiter, line 233: Believe `problemCount` needs to be reset here.

src/App.js
```diff
@@ -113,25 +113,24 @@ const theme = createMuiTheme({
...
-    if (errorCount !== 1) errorString += 's';
-  }
-  if (warningCount > 0) {
-    warningString = `${warningCount} Warning`;
-    if (warningCount !== 1) warningString += 's';
+    if (problemMessages[0] === defaultProblemMessage) {
+      problemMessages = [];
+    }
+    problemMessages.push(message);
+    problemCount++;
```
ngfreiter, line 128: Was the intent to completely sort errors and warnings out of the normal console, or was the problems tab meant to just focus on _only_ the errors and warnings. I'm thinking the console would still include all input, and the problems tab would specifically have the errors. @jafeltra would know better than me, what do you think?

src/components/ConsoleComponent.js
```diff
@@ -24,71 +21,102 @@ const useStyles = makeStyles((theme) => ({
...
   pre: {
     margin: '0px'
+  },
+  circle: {
+    'border-radius': '50%',
+    'min-width': '22px',
+    height: '22px',
+    paddingLeft: 1,
+    paddingRight: 1,
+    background: '#21121d',
```
ngfreiter, line 45: I think this color looks a bit out of place. Consider trying `#263238`. This is the same color as the background of the code mirror component, and I think it looks good.

ngfreiter: Additionally, I think it might work well if there is still some sort of color based indication that something went wrong. Maybe if there are errors, the problem count could be red, and if only warnings, yellow. Similar to how we do it in SUSHI. If you try this and think it looks bad though, feel free to shut this down.

src/components/ConsoleComponent.js
```diff
@@ -24,71 +21,102 @@ const useStyles = makeStyles((theme) => ({
...
+    marginLeft: 4,
+    'justify-content': 'center'
   }
 }));
 
 export default function Console(props) {
   const classes = useStyles();
+  const [problemsView, setProblemsView] = useState(false);
 
   const toggleExpandConsole = () => {
```
ngfreiter, line 57: I think it is best to separate out the expansion button into its own button at this point. The behavior is a bit confusing when the `Console` button and the `^` button are one and the same. For example, let's say I looked at the `Problems` tab, found nothing of concern, and wanted to close the console. You have to double click, see recording:

https://user-images.githubusercontent.com/52747882/134979016-d2553f52-759a-4d9a-b3de-402cc08a0bd5.mov

src/components/ConsoleComponent.js
```diff
@@ -24,71 +21,102 @@ const useStyles = makeStyles((theme) => ({
...
-          <ErrorIcon style={{ display: props.errorCount ? 'block' : 'none' }} className={classes.error} />
-          {props.errorCount ? `${props.errorCount}` : ''}
-          <CheckIcon
-            className={classes.success}
+        </Button>
+        <Button onClick={toggleProblemsConsole} className={classes.problemsButton}>
+          <p
             style={{
-              display: props.consoleMessages.length > 0 && !props.warningCount && !props.errorCount ? 'block' : 'none'
+              'text-decoration': props.expandConsole && problemsView ? 'underline' : 'none',
```
ngfreiter, line 95: I wonder if the underline could be a little lower, more VSCode style:
image

That's just a small stylistic touch though, so feel free to ignore.

src/tests/components/ConsoleComponent.test.js
```diff
@@ -16,41 +16,43 @@ afterEach(() => {
...
 
   const { queryByText } = render(
-    <ConsoleComponent consoleMessages={consoleMessages} errorCount={'2 Errors'} warningCount={'2 Warnings'} />,
+    <ConsoleComponent consoleMessages={infoMessages} problemMessages={problemMessages} />,
     container
   );
-  const warningLabel = queryByText(/2 Warnings/i);
-  const errorLabel = queryByText(/2 Errors/i);
+  // const warningLabel = queryByText(/2 Warnings/i);
+  // const errorLabel = queryByText(/2 Errors/i);
```
ngfreiter, line 52: Is this meant to be removed?

src/tests/components/ConsoleComponent.test.js
```diff
@@ -16,41 +16,43 @@ afterEach(() => {
...
   const successLabel = queryByText(/Success!/i);
 
-  expect(warningLabel).toBeInTheDocument();
-  expect(errorLabel).toBeInTheDocument();
+  // expect(warningLabel).toBeInTheDocument();
+  // expect(errorLabel).toBeInTheDocument();
   expect(successLabel).not.toBeInTheDocument();
```
ngfreiter, line 57: And this?

src/components/ConsoleComponent.js
ngfreiter: For consistency, maybe we could use something like this: 
image

since we already do that on invalid JSON.

ngfreiter (changes requested):
Nice, this definitely will make things easier to navigate. I had a few comments to consider.

src/tests/components/ConsoleComponent.test.js
julianxcarter: That should not have been there 😅. Removed in commit `4388dc0`

src/components/ConsoleComponent.js
julianxcarter: I took this suggestion in commit `4388dc0` and it does look a lot better. Thanks!

src/components/ConsoleComponent.js
julianxcarter: Yeah, this behavior was strange. I thought it made sense for you to be able to open/close the console with either the `Problems` or `Console` buttons, but it makes more sense for each button to do exactly what it says. Commit `4388dc0` separates that little chevron out into its own button so the behavior is more consistent.

src/App.js
julianxcarter: Addressed in `4388dc0`.

src/components/ConsoleComponent.js
julianxcarter: This is a really good idea! I've toyed with the colors a bit in `4388dc0`, adding highlighting based on the error cases that you described (yellow for only warnings, red for errors, white for no errors). Let me know if you think it looks off at all, or perhaps if you think it makes sense to introduce a green color in some way to indicate a successful run.

src/App.js
julianxcarter: The mockup was a bit ambiguous on this, but I do think it makes sense to include _all_ messages in the generic console and then simply include only the problems in the `Problems` console. I've added the error and warning back in commit `4388dc0`.

src/components/ConsoleComponent.js
ngfreiter: I should've been more clear in my initial review, so sorry about that, but what do you think about making it so the `Console` and `Problems` still will open the console to that respective tab? But then if the console is already open, re-pressing the button does not close the tab, and you have to use the chevron to close (same behavior as currently when the console is open). Is that confusing, or does that make sense?

ngfreiter:
Dang this is looking real good to me now, awesome stuff. I had one additional comment on the functionality of `Console` and `Problems` button for you to consider.

src/components/ConsoleComponent.js
julianxcarter: I actually really like that, I don't think it's confusing at all. Just made the change in commit `ec40bbb`.

ngfreiter (changes requested):
Looks great! The one thing I'm confused by is why `package-lock.json` changed. Is it possible that snuck in accidentally somehow? I don't see why it should change.

julianxcarter:
@ngfreiter That `package-lock.json` was definitely accidental and flukey, I just removed it.

cmoesel:
@julianxcarter - I agree that the `package-lock.json` no longer shows up in the diff.  But I am also confused because 41a5ae5, which claims to remove package lock weirdness, seemingly has nothing to do with `package-lock.json`?

julianxcarter:
> @julianxcarter - I agree that the `package-lock.json` no longer shows up in the diff. But I am also confused because [41a5ae5](https://github.com/FSHSchool/FSHOnline/commit/41a5ae5cf557b6f33391cd47bfeb32098393210f), which claims to remove package lock weirdness, seemingly has nothing to do with `package-lock.json`?

Yeah so I rolled back the second to last commit and just removed the package-lock changes from that commit and force pushed it back up. While doing that I noticed that there were some property names that could be streamlined (`'border-color'` to `borderColor` for instance) so I took the opportunity to do that as well 😅

cmoesel:
Ah.  OK.  It will all get squashed in the end anyway.  Approved!

cmoesel (approved):
This is very nice.  Gives the app a little more of a professional flare!  And makes finding those problems nice and easy!

It's outside of the scope of this PR, but I totally think that we should display the summary pun on the bar that has the console and problem tabs -- so I can see my pun without having to open the console.

ngfreiter (approved):

PR #75, opened 2021-09-23
Save individual files and save a zip of all files

jafeltra:
This PR adds the functionality to save each individual file from the action menus in the editors. These buttons work the same way as the delete button and act on the definition in the editor when the button is clicked. There is also a Save All button that saves the FSH file and all JSON files in a zip file.

All files are immediately downloaded to the downloads folder. The FSH file and zip file have very generic names, but the id from the JSON files is used (which is what is displayed in the editor header).


src/App.js
```diff
@@ -309,6 +311,22 @@ export default function App(props) {
     }
   }
 
+  function saveAll() {
+    // Build zip file
+    const zip = new JSZip();
+    zip.file('FSH.fsh', inputFSHText);
+    inputFHIRText.forEach((def) => {
+      const name = `${def.id ?? 'Untitled'}.json`;
```
ngfreiter, line 319: This would be a rare scenario, but if there are two defs without `id`s, they are both saved as `Untitled`, so only one will remain, while others get overwritten.

ngfreiter (changes requested):

src/App.js
jafeltra: Ohh this is an excellent point and I'm not sure I'd even consider it rare (because I often add a bunch of blank definitions). I'll fix that!

jafeltra: This should be resolved in 68e6668.

The edge case that I was testing when writing this can be recreated by:
- Using the FSH [here](https://fshschool.org/FSHOnline/#/share/39HmT5I)
- Adding (at least) three new, blank JSON files
- Adding `{ "id": "Untitled-1" }` to the last blank JSON file you created

This code should keep the provided ids for all JSON files with ids defined, should keep one of the blank files as "Untitled.json", and should append a number of any other "Untitled" JSON files that does not conflict with an existing file name.

ngfreiter (approved):
Nice, looks good to me 👍

julianxcarter (commented):
This looks good to me! The only suggestion that I have is that perhaps we should put the resourceType before the actual id of each resource in the fileName (e.g. `StructureDefintion-example-profile.json`). That's what SUSHI does, so we may want to do it on FSHOnline just for consistency's sake. If other's don't think this is necessary, then this has my second approval!

cmoesel:
I think @julianxcarter makes a great point and would advocate for using the same file-naming scheme as SUSHI uses when it generates FHIR files.  (Which, as Julian said, is `${type}-${id}.json`.

ngfreiter:
@julianxcarter, I made some updates to your `resourceType` file saving in f582dfc. I was having some issues when `resourceType`/`id` were not actually defined. Also I noticed that confusingly, `def.resourceType` is not the actual `resourceType` of the resource.

julianxcarter (approved):
Looks good to me!

PR #76, opened 2021-10-12
Update Meta Tags

jafeltra:
This PR makes minor changes to the `<meta>` tags we use. The main noticeable difference is that I added the image that gets displayed when you send a link. I've been testing this out by sending links over slack. I don't know how to test this locally, so I pushed the same code that's in this PR to my forked version of this repo and deployed it to my GitHub pages. You can test this PR by sending this link: https://jafeltra.github.io/FSHOnline/#/. The URL should also expand in Slack if you send a Share link as well.

There are a few other things I think could potentially be implemented, but I'm not sure if the information we would add would provide that much value, so I'm not sure it's worth implementing them now. These are a few of the links I sent myself in slack to see what the URLs looked like:

- https://github.com/FHIR/sushi
- https://www.npmjs.com/package/fsh-sushi
- https://codepen.io/jafeltra/pen/XWarxEq

If any of those have attributes that look particularly exciting, let me know and I can try to add them!


julianxcarter (approved):
I'm hooked!

PR #78, opened 2021-11-17
Open Console when it is closed and a problem occurs

jafeltra:
This PR make it so the Console automatically opens if it is not already open when a problem occurs. It can still be collapsed and reopened after it initially opens itself. It only opens if a problem occurs - if there are no issues logged, it will stay closed if it is already closed.

### DIFF
--- a/src/components/ConsoleComponent.js
+++ b/src/components/ConsoleComponent.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { makeStyles } from '@material-ui/core/styles';
 import { Box, Button, IconButton } from '@material-ui/core';
 import { ExpandLess, ExpandMore } from '@material-ui/icons';
@@ -59,6 +59,16 @@ const useStyles = makeStyles((theme) => ({
 export default function Console(props) {
   const classes = useStyles();
   const [problemsView, setProblemsView] = useState(false);
+  const { problemCount, setExpandConsole } = props;
+
+  useEffect(() => {
+    // When props.problemCount changes and if there is a problem, make sure to open the console.
+    // If the problemCount is 0, we don't need to open it.
+    // The console should be able to be closed, so don't track whether the props.expandConsole value has changed.
+    if (problemCount) {
+      setExpandConsole(true);
+    }
+  }, [problemCount, setExpandConsole]);
 
   const toggleExpandConsole = () => {
     setProblemsView(false);
--- a/src/components/ConsoleComponent.js
+++ b/src/components/ConsoleComponent.js
@@ -71,7 +71,6 @@ export default function Console(props) {
   }, [problemCount, setExpandConsole]);
 
   const toggleExpandConsole = () => {
-    setProblemsView(false);
     props.setExpandConsole(!props.expandConsole);
   };
 


src/components/ConsoleComponent.js
```diff
@@ -59,6 +59,16 @@ const useStyles = makeStyles((theme) => ({
...
   const classes = useStyles();
   const [problemsView, setProblemsView] = useState(false);
+  const { problemCount, setExpandConsole } = props;
+
+  useEffect(() => {
+    // When props.problemCount changes and if there is a problem, make sure to open the console.
+    // If the problemCount is 0, we don't need to open it.
+    // The console should be able to be closed, so don't track whether the props.expandConsole value has changed.
+    if (problemCount) {
+      setExpandConsole(true);
```
julianxcarter, line 69: Do you think we should set the console to open to the problems view when this happens? On the one hand, I think that extra bit of red conveys that something went wrong here. On the other, there may be some messages logged to the main console that may help to diagnose the problem further.

julianxcarter (commented):
Looks good! Just one thought.

src/components/ConsoleComponent.js
jafeltra: That's a good question. If we did add that functionality, does that mean it should swap over to the problems Problems tab when there is an error but the console was already open to the full Console tab? I think the way it is now means that whichever tab you already have open will stay open any time you use one of the "convert" buttons if the console is open. It feels a little odd to swap to the Problems tab if the Console tab is open and you are looking for messages there.

I'm thinking that with that piece and the fact that you do get the additional context, maybe it's not worth adding in. But also open to your opinion too!

julianxcarter: Yeah I hadn't thought about that, having the console switch to the other view if it's already open would feel really strange. I think this is the best way to handle that!

julianxcarter (approved):
Looks good!

src/components/ConsoleComponent.js
cmoesel: How hard would it be to open to the problems tab if the console is collapsed, but leave it as-is if the console is already expanded?  If not difficult, what do you think about that approach?

jafeltra: I was just looking into that. I'll report back.

cmoesel:
It seems to work as described -- and I _do_ think that auto-expanding it does make it more obvious that something went wrong.  So that's good!

As discussed in one of the specific comment threads, I think it would make sense to open to the problems tab if the console is collapsed.  If it's already open to the console though, I wouldn't switch it; I'd leave it as-is.  (Also, I noticed that even if I switch to the Problems tab and then collapse it, when it auto-opens, it doesn't remember I was on the Problems tab; it just goes to the Console -- which is kind of like what we want, but the exact opposite ;-).

src/components/ConsoleComponent.js
jafeltra: I remembered why this was harder than I thought. At least the way I have it set up with the `useEffect`, there isn't a way to tell if the console is closed because it was already closed and it should be opening on the "convert" buttons running, or if it's closed because we just clicked the collapse button and we want it to stay closed. There might be a way to move where the function is called, but I'm not sure what the right place for it should be.

jafeltra: The only other place I can think to put this functionality is within the function that sets the messages themselves. But since that is just attached to the `console.log` function we overwrite, it isn't really tied to any of the state functions we use to open and close the console. I might be over-complicating it, but I can't figure out how to piece those two things together. If we ever decide to color code the console messages (like we get in a terminal for SUSHI), we might need to update how that function works, and maybe we can update this functionality then.

But also, if I'm over complicating things and either of you see an easy solution I'm missing, please let me know!

jafeltra:
>As discussed in one of the specific comment threads, I think it would make sense to open to the problems tab if the console is collapsed. If it's already open to the console though, I wouldn't switch it; I'd leave it as-is. 

I mentioned this on the thread, I can't figure out a nice way to do this. Because the messages itself just get logged through our overwritten `console.log` function, I can't figure out a nice way to tie the React state to this function and appropriately open the right tab at the right time. I can keep looking into it if this is something we think is worthwhile, and it might be! I just wanted to check before spending more time on it.

>(Also, I noticed that even if I switch to the Problems tab and then collapse it, when it auto-opens, it doesn't remember I was on the Problems tab; it just goes to the Console -- which is kind of like what we want, but the exact opposite ;-).

I did fix this at least! Now when you expand the console, it stays on whatever tab you were previously on. So if you had the Problems tab open, it should stay on that, and if you had the Console tab open, it should stay on that! That is in 5df988d.

cmoesel:
> I can keep looking into it if this is something we think is worthwhile, and it might be! I just wanted to check before spending more time on it.

Nah.  If there isn't an easy way forward, I think we can keep this on the backburner for now.

cmoesel (approved):
Looks good.  Works well.  Thanks!

PR #83, opened 2021-12-14
Filter #inline Instances

### LINKED ISSUE #79: Instances of datatypes (not resources) are listed as "Untitled" in "Unknown Type"
For example, an instance of a Quantity shows up in the UI as "Untitled" in "Unknown Type".  See this example:
https://fshschool.org/FSHOnline/#/share/3cFdPQm

I expect this is because the JSON doesn't have a `resourceType` property.

jafeltra:
This PR updates FSH Online to more closely mirror what SUSHI does with `#inline` instances. The instances are now filtered out of the package that is then rendered on the RHS of the app. Inline instances will no longer appear as their own standalone JSON, but instead will just be used inline with the resource they are used within.

This also adds a new `logger.info` message with the total count of resources that are exported to JSON. This also mirrors the logged messages that SUSHI outputs as it writes only resources to files. This count will be different from the one in the results box, but that also mirrors what SUSHI displays.

This adds a test for this case and running the provided example ([here](https://fshschool.org/FSHOnline/#/share/3cFdPQm)) on this branch should remove the Quantity from the JSON displayed on the right.

Fixes #79


mint-thompson (approved):
no more unknown type files in the export! hooray!

julianxcarter (approved):
Looks good!

julianxcarter (approved):

PR #84, opened 2021-12-21
Update SUSHI to v2.2.4 and GoFSH to v1.4.2

jafeltra:
Update to latest versions.


cmoesel (approved):
This looks good! I tested it by clearing my browser storage (to ensure dependency download still works) and trying FSH that exercised a bug that was fixed in SUSHI 2.2.6.  As expected, the conversion worked (meaning it _is_ using SUSHI 2.2.6).  APPROVED!

PR #87, opened 2022-01-18
Catch error from http.get and skip tests

jafeltra:
This PR adds proper error handling if the `http.get` call fails. This was happening when `https://packages.fhir.org` had the invalid cert error. When this happened, the app would throw an error that was not handled. This caused the indexedDB in the browser to be set up incorrectly, so it could never later successfully download the same package it failed on. It also caused the loading spinner in the "To JSON" button to never stop spinning.

This PR catches the error, logs an error, and resolves the promise with the typical "empty" set of packages we use in other places. This means that the app no longer throws an error, it properly cleans up indexedDB, the spinner stops, and the console pops open with the error message. It also means we can properly download the package later on. The majority of these changes are whitespace changes.

To test this locally with this branch, since `https://packages.fhir.org` is working again, I've been testing with a URL that doesn't exist, like `https://test.packages.fhir.org`. On master, if you clear out your indexedDB database, you'll get an error logged in the web console and the app will look like it is hanging. On this branch, the app's console will open with an error message. You should also be able to replace the original url and the packages should download as expected.

I have no idea how to test this. If anyone reviewing this has some brilliant insight into it, please save me from myself. Otherwise, the tests are skipped.


cmoesel:
Those tests were tricky! I finally got them to work by introducing `nock` and mocking out a request that actually returns a very small `.tgz` file (so the rest of the unzip code doesn't fail).

In the process of doing so, I ran into two issues.  I don't know why we did not see this before, but apparently if you are issuing requests to an https site, you are supposed to use the `https` module (not `http`).  I discovered this because when I got `nock` working, I started getting the error discussed in [this StackOverflow question](https://stackoverflow.com/questions/34147372/node-js-error-protocol-https-not-supported-expected-http).  The other issue I fixed, after discussing with @jafeltra, was to add the failed package to `emptyDependencies` after the error is caught.

cmoesel (approved):
I've tested this and it all seems to be working well, but since I have now authored some of this code, we definitely want another reviewer on it!

jafeltra:
Ugh. Wow. Thank you for your brilliance and patience in figuring this out. I definitely feel much better knowing there is a way to test the functionality we actually want.

And thanks for catching that missing line! I think I must have had it on some stash and completely forgot that I had changed it. I approve of your changes/my own PR! Woo!

julianxcarter (approved):
Approved!

PR #103, opened 2022-06-30
Update GoFSH to v1.6.1

jafeltra:
This updates FSH Online to use the latest release of GoFSH. I tested this with [GoFSH #186](https://github.com/FHIR/GoFSH/issues/186) and confirmed that the latest version of GoFSH works in FSH Online and that the bug that is shown there and fixed in 1.6.1 is properly created in FSH Online.


cmoesel (approved):
Yay!

PR #105, opened 2022-08-25
Update SUSHI and GoFSH

jafeltra:
Update SUSHI to v2.6.1 and GoFSH to v1.6.2


cmoesel (approved):
Hurray for progress.  The wheel keeps on turning.

PR #120, opened 2022-12-21
Add GoFSH indent option

mint-thompson:
Completes task [CIMPL-977](https://standardhealthrecord.atlassian.net/browse/CIMPL-977).

A checkbox is provided in the configuration menu to control whether the FSH produced by GoFSH is indented. The value of this checkbox is not stored in the shared configuration included in generated URLs. The checkbox is unchecked by default.


cmoesel (changes requested):
I'm not a React expert, but the code looks reasonable to me.

I've tested it and it works well. My only suggestion is to use "Convert to FSH" instead of "GoFSH"in the config UI.

jafeltra (changes requested):
This looks good! I had a few small comments inline, but I also have one more that I couldn't comment on inline:

- The `package-lock.json` was updated to use the `lockfileVersion` 2. I'm not sure if there are actually any negative consequences for doing this, but we might want to ensure that there aren't. Do you mind reverting the `package-lock.json` file until we do fully test that?

jafeltra (changes requested):
This looks good! The only remaining thing I had noticed is the changes in `package-lock.json`:

The `package-lock.json` was updated to use the lockfileVersion 2. I'm not sure if there are actually any negative consequences for doing this, but we might want to ensure that there aren't. Do you mind reverting the package-lock.json file until we do fully test that?

mint-thompson:
Based on my understanding of the lockfile documentation, version 2 should be fully backwards compatible, but just to keep things simple, I've reverted the file.

jafeltra (approved):
Thanks @mint-thompson! I was thinking that there might be updates to dependencies that we didn't test out, but I think you're right that if it's backwards compatible, it probably wouldn't make a difference. Maybe we can update it soon! Or if you feel strongly, we can add it back here.

cmoesel:
Thanks @mint-thompson! As I was testing this, I noticed that it is not using the nice indentation improvements that Guhan and I worked on.  It turns out, that is because we have not released GoFSH since then.  I think maybe we should release GoFSH and integrate it into this PR before merging this and making it live.  I'd like people's first experience to reflect the best indentation we have to offer.

Is that okay with you?

cmoesel (approved):
OK! Now we have super fancy indenting! Excellent!

image

_(If you know, you know)_

cmoesel:
@mint-thompson approved the PR that got merged into their branch -- so I think that's basically and approval of the changes that were made to their branch. So I say it's good to go!

PR #124, opened 2023-02-09
Update SUSHI, GoFSH, and other dependencies

jafeltra:
This PR updates SUSHI and GoFSH to their latest version. I also ran an `npm audit fix` to clean up some of the dependencies. This won't resolve all the dependabot PRs because there are some issues that require updating `react-scripts` to the latest major version, which I didn't want to do within this PR, but it should close some of them.

I had this branch target #120 so that the indent option could be tested with the latest GoFSH, and so that PR didn't need to get updated again. The only thing I had to change was updating the indentation that is output with the `--indent` flag in a test. If it makes more sense to go into `master` first, just let me know.


mint-thompson (approved):

cmoesel (approved):
This looks good to me. I confirmed that it works as expected, and I agree w/ merging it into the `cimpl-977-gofsh-indent` branch.

image

PR #142, opened 2023-09-14
Add a link to the FSH Online Examples repo

jafeltra:
Add a short blurb with a link to the repo that holds the examples. Previously, this was only in the README file, but it would be useful to have it in the app itself.

Two small pieces to consider when reviewing this:

- Is the text too prominent? I could make it smaller and/or lighter, but I didn't want to get unnecessarily fancy. Let me know if you'd like screenshots of these options to consider. _**Edit**: I decided to make the font size "small" so now this question is better posed as - is it too small? not prominent enough?_
- How do you feel about the wording? It's a little joke-y, so I could definitely simplify it with something just like "View and add to the examples collection on Github".

### DIFF
--- a/src/components/FSHControls.js
+++ b/src/components/FSHControls.js
@@ -23,6 +23,7 @@ import {
   FormControlLabel,
   FormHelperText,
   Grid,
+  Link,
   MenuItem,
   Tooltip,
   TextField
@@ -89,6 +90,13 @@ const useStyles = makeStyles((theme) => ({
   dialogPaper: {
     maxHeight: '80vh',
     minHeight: '80vh'
+  },
+  dialogActions: {
+    justifyContent: 'space-between'
+  },
+  dialogActionsMessage: {
+    fontStyle: 'italic',
+    padding: '6px 8px'
   }
 }));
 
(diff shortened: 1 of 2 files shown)


cmoesel (changes requested):
Thanks for this. I offered my opinion on the font size and made one suggestion regarding the link.

BTW - I didn't find the text too joke-y. After you warned us, I was expecting something along the lines of: _Not enough FSH in the sea? Contribute..._

jafeltra:
Okay, I think @cmoesel  is right and I didn't use a pun but I should have. Here are some options. Let me know if anyone likes one or a different combo of various fish-related puns:

> Sea an example that is missing? Lend a kelp-ing hand and add it on GitHub!

> Have an example that might be bene-fish-al? Seas the day and add to our collection on GitHub!

or from @cmoesel:

> Not enough fish in the sea? Contribute an example on GitHub!

cmoesel:
Oh, you gotta go for the double pun. Twice the fun!
> Have an example that might be bene-fish-al? Seas the day and add to our collection on GitHub!

You think the joke's over and then BAM! -- there's another one!

cmoesel (approved):
Love it.

mint-thompson (approved):

PR #147, opened 2023-09-22
Update Node to Node 18

cmoesel:
Update the .tool-versions and CI scripts to Node 18. This also requires a temporary workaround to avoid SSL errors when running or building the app (`NODE_OPTIONS=--openssl-legacy-provider`). According to the interwebs, we can back out that workaround  once we update react-scripts to v5 or greater.


jafeltra (dismissed):
Hopefully we'll be able to remove the workaround soon, which is good. Besides that, this looks good!

mint-thompson (dismissed):

PR #149, opened 2024-01-11
SUSHI and GoFSH version updates, minor version updates

jafeltra:
This PR updates minor and patch versions of dependencies where possible.

It also updates to the latest versions of SUSHI and GoFSH. In order to get them to work in FSH Online, I had to use the latest version of `antlr4`, so I added an `override` in `package.json` to override the `antlr4` versions specified by both SUSHI and GoFSH. Using the latest version let us take advantage of a fix to the package entrypoints. I did locally test out updating `antlr4` to this new version in both SUSHI and GoFSH, and nothing seemed to break, so I think this override is safe to use. Once a released version of SUSHI and GoFSH each use the new version, the override can be removed.

Also, with the fixes in the newer `antlr4` version, the Jest resolver we previously needed is no longer needed, so I removed it. While there were some useful links in the file, it will live on in version history, so I got rid of it.


mint-thompson (approved):
Tested this out locally and everything appears to be in good working order. Hooray for updates!!

PR #150, opened 2024-01-23
Update to SUSHI 3.6.1 and GoFSH 2.3.0

jafeltra:
Updates to use the latest SUSHI and GoFSH.

I tested out the same Code System that caused an issue between the previous SUSHI and GoFSH versions, and things seem to be corrected.


cmoesel (approved):
This looks good! I've also tested it out locally and confirmed it works as expected (and fixes the bug).

PR #151, opened 2024-02-02
Update display for resources without id or resourceType

jafeltra:
This PR improves handling for examples of Logical Models (or any JSON) that does not have an `id`. If there is no `id`, we try to calculate a related "display" to use in the sidebar on the right and on the top of the editor next to "FHIR JSON:". If there is no `id` or `resourceType`, then we'll fall back to the usual 'Untitled'.

To test this, you can try out using the new example of logical models that was recently contributed on this branch: [here](https://fshschool.org/FSHOnline/#/share/4botLUg).

### DIFF
--- a/src/components/JSONOutput.js
+++ b/src/components/JSONOutput.js
@@ -114,6 +114,18 @@ const checkFshType = (def) => {
   return null;
 };
 
+const checkIdToDisplay = (def) => {
+  if (def.id) {
+    return def.id;
+  }
+  if (def.resourceType) {
+    // Logical Models don't always have an id, so fall back to last part of the resourceType
+    return def.resourceType.substring(def.resourceType.lastIndexOf('/') + 1);
+  }
+  // If can't determine an id or a best guess from the resourceType, just return 'Untitled'
+  return 'Untitled';
+};
+
 // Flatten the package so we can render and navigate it more easily,
 // but keep high level attributes we'll need accessible
 const getIterablePackage = (defsPackage) => {
@@ -126,7 +138,11 @@ const getIterablePackage = (defsPackage) => {
     ...defsPackage.valueSets,
     ...defsPackage.codeSystems
   ];
-  return defArray.map((def) => ({ resourceType: checkFshType(def), id: def.id, def: JSON.stringify(def, null, 2) }));
+  return defArray.map((def) => ({
+    resourceType: checkFshType(def),
+    id: checkIdToDisplay(def),
+    def: JSON.stringify(def, null, 2)
+  }));
 };
 
 export default function JSONOutput(props) {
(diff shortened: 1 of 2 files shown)


cmoesel (commented):
This works. I think my only comment is that although I was the one who recommended just taking the last part of the `resourceType`, I wonder if we should also prepend it with `Instance Of` or append it w/ `Instance`. I was a little surprised when I ran the example and saw `Aquarium` twice on the right-hand side -- and then I realized why. But maybe `Instance of Aquarium` or `Aquarium Instance` would have been better.

The most ideal solution, however, would be to use the name that the author used in the FSH (which was `ExampleAquarium` here). Is there any way to tie back to that or is it all too abstracted away from us to reconnect it?

jafeltra:
> I was a little surprised when I ran the example and saw `Aquarium` twice on the right-hand side -- and then I realized why. But maybe `Instance of Aquarium` or `Aquarium Instance` would have been better.

I also felt this way when I first saw it, so knowing you did too, maybe it is worth switching to something like this. I convinced myself it was okay because one "Aquarium" should be in the "Instances" section and the other should be in the "Logical Models" section. But you're right that it is a little jarring, so maybe I can figure out something better.

> The most ideal solution, however, would be to use the name that the author used in the FSH (which was ExampleAquarium here). Is there any way to tie back to that or is it all too abstracted away from us to reconnect it?

The issue with this approach is that when we're displaying all the FHIR JSON output on the right hand side, we only know about the FHIR resources that SUSHI outputs. The `ExampleAquarium` name here is only used in the FSH and never translated into the FHIR resource at all. So there isn't a straightforward way to know the FSH names or re-associate them.

I can look into some options to make the name clearer. Maybe there's something I'm missing.

jafeltra:
> I can look into some options to make the name clearer. Maybe there's something I'm missing.

@cmoesel - I looked into this for a bit and realized that the `_instanceMeta` from SUSHI is helpful here and has what we really want (`ExampleAquarium`). However, we use `JSON.stringify` in FSH Online to display the instances in the editor, which uses the `toJSON` from `InstanceExporter`, which strips out the `_instanceMeta`. I was able to update how `JSON.stringify` handles instances to avoid doing this, but it is not trivial to keep the metadata around where we need it without displaying it in the editor. Before I spent more time trying to find the right way to pass along this info, I wanted to check in with you on what you think about alternate displays, like `Aquarium Instance`.

jafeltra:
After chatting about it offline, [02f7353](https://github.com/FSHSchool/FSHOnline/pull/151/commits/02f7353dabc8ab80cddf8ddbdc9c8f022d0bb22e) updates this PR to say "Instance of <resourceType>". This hopefully makes it clearer that the label we use isn't an `id` but gives the most information we can without refactoring the component right now. This case should really only come up from SUSHI when you have an instance of a Logical Model that doesn't have an `id` element. It could also come up if someone was hand typing some JSON or deleted the relevant properties from a full JSON. But I think the current way it is handled is still accurate for those cases.

cmoesel (approved):
Looks good to me and lines up w/ our previous decision as a team. Thanks!

mint-thompson (approved):

PR #153, opened 2024-02-21
Update automatically loaded dependencies

jafeltra:
This PR updates FSH Online to automatically load dependencies that are more in line with SUSHI's current implementation. We use the `latest` version of `hl7.fhir.uv.tools` (rather than `current`, like SUSHI) because FSH Online does not support `current` versions at the moment. The `hl7.terminology` and `hl7.fhir.uv.extensions` packages are now loaded based on the current FHIR version being used, similar to SUSHI.

I made a small refactor to the code in the second commit because it seemed simpler.

You can test this by changing to the different FHIR versions FSH Online supports and ensuring that the proper packages are loaded for each one.

### DIFF
--- a/src/utils/FSHHelpers.js
+++ b/src/utils/FSHHelpers.js
@@ -242,16 +242,32 @@ function addCoreFHIRVersionAndAutomaticDependencies(dependencies, coreFHIRVersio
   if (!hasCoreFHIR) {
     dependenciesToAdd.push(coreFHIRPackage);
   }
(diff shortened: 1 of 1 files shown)


cmoesel (changes requested):
Thanks for doing this. There is a little squishiness with R4/R4B that works now but could cause problems in the future if new automatic dependencies are added with specific R4 or R4B packages. I've also suggested a couple of other possible refactorings to consider. Since we don't support as many FHIR versions as SUSHI does, I think we can have a simpler approach than SUSHI's.

cmoesel (approved):
Hurray for free packages!

PR #155, opened 2024-03-22
Add variations on start and build scripts for Windows systems

mint-thompson:
These scripts set an environment variable before invoking react-app-rewired. The syntax for doing this is different on Windows systems. Two new scripts are defined to make these actions available to Windows users.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,6 +32,7 @@
       },
       "devDependencies": {
         "cors": "^2.8.5",
+        "cross-env": "^7.0.3",
         "eslint-config-prettier": "^6.11.0",
         "eslint-plugin-prettier": "^3.1.3",
         "express": "^4.18.2",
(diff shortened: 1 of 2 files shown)


jafeltra (dismissed):

jafeltra:
I tried using `start:win` on my Mac and it does not work. So I guess we can discuss whether we want to add cross-env and use that in one script or keep both.

cmoesel:
I say let's use [cross-env](https://www.npmjs.com/package/cross-env). It's designed exactly for this purpose and is super easy to use.

cmoesel:
I had some time, so I switched this over to using `cross-env`. It still works on Mac. At this point, we can only assume it works on Windows (if there is truth in advertising).

cmoesel (approved):
Wow, Mint and Chris. This looks great!

jafeltra (approved):

PR #156, opened 2024-04-16
Update to SUSHI 3.9.0 and GoFSH 2.3.1

jafeltra:
Updates to the latest SUSHI and GoFSH versions.

I tested out new features from both, and everything worked as expected. I did have to add the same override we needed for SUSHI now that we're using a newer GoFSH.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,8 +19,8 @@
         "browserify-zlib": "^0.2.0",
         "codemirror": "^5.65.16",
         "file-saver": "^2.0.5",
-        "fsh-sushi": "^3.8.0",
-        "gofsh": "^2.3.0",
+        "fsh-sushi": "^3.9.0",
+        "gofsh": "^2.3.1",
         "jszip": "^3.10.1",
         "lodash": "^4.17.21",
         "react": "^16.13.1",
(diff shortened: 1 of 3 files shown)


cmoesel (changes requested):
When I run `npm install`, I get changes in my `package-lock.json`. They seem to be related to `https-proxy-agent` in `fhir-package-loader`.

When I run `npm run build`, the build fails:
> Failed to compile.
>
> ./node_modules/fhir-package-loader/node_modules/https-proxy-agent/dist/index.js 70:29
> Module parse failed: Unexpected token (70:29)

It seems like maybe you need to add that override for `fhir-package-loader` as well?

jafeltra:
Oh wow, I'm not sure how I missed that. Thanks for catching it!

jafeltra:
It turns out that `gofsh` doesn't actually directly include `https-proxy-agent` and just uses it through `fhir-package-loader`. I switched the override to be `fhir-package-loader` and that seemed to fix things. I also don't get any more changes locally in `package-lock.json`. Let me know if both of those issues are resolved for you as well.

cmoesel (approved):
It's good!

PR #158, opened 2024-05-31
Migrate from Create React App to Vite

jafeltra:
This PR removes the Create React App framework and migrates to the [Vite](https://vitejs.dev/) framework. This will (hopefully) make maintenance going forward easier to manage with a lighter weight, currently maintained framework. There were quite a few things that changed and I tried to keep them in relatively separate commits, but I've included a brief summary of the changes here:

- Moved `AppRouter` out of `utils` since it didn't really make sense as a util
- Switched all source and test files with React component to be `.jsx` files
- Removed `react-scripts`, `react-app-rewired`, and the serviceWorker code since they are CRA specific
- Added and configured a `vite.config.mjs` file
- Moved `index.html` and updated it for Vite
- Renamed env vars to have VITE_ prefix and used `import.meta` in place of `process.env`
- Stubbed out packages that are file system specific that FSH Online does not need and fhir/fhir because it is used by SUSHI but will never be needed in FSH Online
- Updated ESLint configuration to no longer rely on CRA
- Switched to React 17
- Switched to Node 20
- Switched from Jest (included with react-scripts) to Vitest
- Updated the README to remove references to CRA
- Removed the `bitly` dependency and just make the two calls we need with `fetch` directly

To test this, you should
- `npm install` (it might be good to first delete `node_modules` if you have it to start fresh)
  - I regenerated the `package-lock.json` a few times to make sure it was up to date, so make sure there are no changes after you do an install
- use `npm start` to run the development server
- test the app to make sure it works as expected
  - Try using SUSHI, using GoFSH, loading a Share link, saving FSH/JSON/all files, and using the Examples.
  - Also, remove any saved definitions from your browser's storage in IndexedDB and then try to run SUSHI/GoFSH to ensure packages can still be loaded properly
- use `npm run build` to build a static, production build (this will be in `dist` now, not `build`, since that's the Vite default)
- use `npm run preview` to serve the static build and test this out in the same way as above
- use `npm test` to make sure the tests run and pass
- use `npm run lint` and `npm run prettier` to ensure the updated configuration works as expected

Two other things to note:
- To get the "format on save" to work in VS Code, I did have to add a File Association in my editor settings. Let me know if you run into the same thing and I can help.
- I think I used the `base` property in the `vite.config.mjs` file properly so that the deployment will work, but if anyone notices anything that might specifically be off there, please let me know.


.github/workflows/deploy-workflow.yaml
```diff
@@ -13,11 +13,11 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
         with:
-          node-version: '18'
+          node-version: '20'
       - run: npm install
       - run: npm run build
         env:
-          REACT_APP_BITLY_KEY: ${{ secrets.REACT_APP_BITLY_KEY }}
+          VITE_BITLY_KEY: ${{ secrets.VITE_BITLY_KEY }}
```
cmoesel, line 20: I think this means we'll need to remember to update the SECRETS in the GitHub settings too, right?

.github/workflows/deploy-workflow.yaml
```diff
@@ -13,11 +13,11 @@ jobs:
...
       - name: Publish build to github pages
         uses: peaceiris/actions-gh-pages@v3.5.7
         with:
```
cmoesel, line 23: It won't let me comment on line 25 below, so this is the closest I can get. Anyway, since Vite builds to `dist` and not `build`, I think we need to change `publish_dir` from `build` to `dist`.

index.html
```diff
@@ -2,44 +2,36 @@
...
 
     <meta name="twitter:title" content="FSH Online"/>
     <meta name="twitter:description" content="A web playground to write and share FHIR Shorthand. It's so fun, you're guaranteed to get hooked!"/>
     <meta property="twitter:url" content="/FSHOnline"/>
     <meta name="twitter:card" content="summary"/>
-    <meta name="twitter:image" content="%PUBLIC_URL%/FSH-online-logo.png"/>
+    <meta name="twitter:image" content="/FSH-online-logo.png"/>
 
-    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
+    <link rel="apple-touch-icon" href="/favicon.ico" />
```
cmoesel, line 21: Was this just wrong before?

src/components/CodeMirrorComponent.jsx
```diff
@@ -13,15 +13,15 @@ import 'codemirror/lib/codemirror.css';
...
-require('codemirror/addon/scroll/simplescrollbars');
+import 'codemirror/addon/mode/simple';
+import 'codemirror/addon/edit/closebrackets';
+import 'codemirror/addon/display/placeholder';
+import 'codemirror/addon/comment/comment';
+import 'codemirror/addon/fold/foldgutter';
+import 'codemirror/addon/fold/brace-fold';
+import 'codemirror/mode/xml/xml';
+import 'codemirror/mode/javascript/javascript';
+import 'codemirror/addon/scroll/simplescrollbars';
```
cmoesel, line 24: Weird. Did we use `require` for a _reason_ before? Or was it just an oversight?

src/components/JSONOutput.jsx
```diff
@@ -305,7 +305,7 @@ export default function JSONOutput(props) {
       new Blob([def]),
       resourceObj?.resourceType && resourceObj?.id
         ? `${resourceObj.resourceType}-${resourceObj.id}.json`
-        : `${name}.json`
+        : `${defaultName}.json`
```
cmoesel, line 308: Whoops. Good catch!

src/tests/components/ShareLink.test.jsx
```diff
@@ -4,20 +4,26 @@ import { unmountComponentAtNode } from 'react-dom';
...
 import * as bitlyWorker from '../../utils/BitlyWorker';
-import Zlib from 'browserify-zlib';
 
 // Mock copy to clipboard since we don't need to test the component itself
-jest.mock('copy-to-clipboard', () => {
-  return jest.fn();
+vi.mock('copy-to-clipboard', () => ({
+  default: vi.fn()
+}));
+const { deflateSpy } = vi.hoisted(() => {
```
cmoesel, line 12: `vi.hoisted`! Intriguing!

src/tests/utils/Load.test.js
```diff
@@ -17,7 +17,7 @@ describe('#unzipDependencies', () => {
   });
 
   it('should make an http request and extract data from the resulting zip file', async () => {
-    const scope = nock('https://packages.fhir.org')
+    const scope = nock('http://packages.fhir.org')
```
cmoesel, line 20: So what's this all about?

src/utils/BitlyWorker.js
```diff
@@ -1,36 +1,30 @@
...
-          console.error(
-            'Error: REACT_APP_BITLY_KEY needs to be set as an environment variable in order to share FSH Online links.'
-          );
-          resolve({ link: undefined, errorNeeded: true });
-        } else {
-          resolve({ link: undefined, errorNeeded: true });
-        }
-      });
-  });
+  return fetch('https://api-ssl.bitly.com/v4/shorten', {
```
cmoesel, line 2: Just curious -- was the `bitly` library causing issues? Not maintained? Or did you just decide that bitly is easy enough that we don't need another dependency just for bitly?

cmoesel (changes requested):
I have not tested this yet, but I did a sweep through the code changes. Everything looks pretty awesome. I like it!

I did find one thing that I'm pretty sure needs changing (the deploy action). I also left a few comments and questions -- many of which probably don't require changes.

Thanks for working this all out. I am personally convinced that it will be easier to maintain in the future. So thank you.

.github/workflows/deploy-workflow.yaml
jafeltra: Yep. I already added a new secret to Github. I can remove the old one once this is deployed since it won't be needed anymore.

.github/workflows/deploy-workflow.yaml
jafeltra: Oh good catch. I'll update this.

index.html
jafeltra: Yeah it was. You can see this even on the currently deployed version. There's a 404 error because it can't find that file. We deleted that file, which was a React logo added by CRA, but we never updated the reference to it. I believe the `apple-touch-icon` is used as the app cover, so I don't think it ever affected anything on the browser, so it hadn't really mattered, but I just cleaned it up while I was in the file.

src/components/CodeMirrorComponent.jsx
jafeltra: I'm not sure why we had `require` before. Maybe that's what it had in the documentation and we just mimicked it? But Vite complained about it, and it seemed to still work this way, so I changed it.

src/tests/components/ShareLink.test.jsx
jafeltra: This was the only way I could figure out how to mock `browserify-zlib` and have access to the mocked function to check what it was called with. Because the vi.mock gets lifted to the top of the file, I had to use the vi.hoisted to make sure the `deflateSpy` mock was available before the vi.mock is created. If you know of a better way to do this, I'm happy to change it.

src/tests/utils/Load.test.js
jafeltra: This is what happened when I mocked `https` in the `nodePolyfills` plugin. I really can't explain why things broke or why this fixed it.

src/utils/BitlyWorker.js
jafeltra: All of the above actually. The `bitly` library was using an old version of axios that was reporting an issue in `npm audit`. When I looked at the library, it hadn't been updated in a while and the calls were easy enough when I looked at the documentation, so I figured it was worth updating to remove the dependency.

.github/workflows/deploy-workflow.yaml
jafeltra: Updated in [94697af](https://github.com/FSHSchool/FSHOnline/pull/158/commits/94697af9b14905d2262bd658391a8b7d819c2b16).

cmoesel (changes requested):
The changes looks good.

This works great when I run `npm start`.

When I run `npm run build`, followed by `npm run preview`, however, it doesn't work so well. When I load it in my browser, I get a blank white screen and the following error in my browser console:
```
vendor-DGUO_56M.js:905 Uncaught TypeError: Object.defineProperty called on non-object
    at Function.defineProperty (<anonymous>)
    at p5t (vendor-DGUO_56M.js:905:8567)
    at vendor-DGUO_56M.js:906:1693
    at oW (vendor-DGUO_56M.js:906:1710)
    at vendor-DGUO_56M.js:646:1463
    at n3t (vendor-DGUO_56M.js:711:1483)
    at vendor-DGUO_56M.js:711:5258
    at yl (vendor-DGUO_56M.js:711:5294)
    at vendor-DGUO_56M.js:711:5473
    at vendor-DGUO_56M.js:712:6144
```

Looking at the first location in the stack, it's trying define `__esModule` on something -- but it's hard to tell what that something is due to code minimization. It seems to be somewhere in the package loading code.

CleanShot 2024-06-07 at 15 00 55

jafeltra:
@cmoesel - thanks for noticing the build issue. I was able to reproduce it, but I wasn't able to reliable reproduce it. I could run the `build` and `preview` tasks, and it would break only about 1 out of 3 times. I tried a few different ways around this, and the only way I've found so far is defining the variable that is sometimes undefined globally in the configuration ([d9c06c8](https://github.com/FSHSchool/FSHOnline/pull/158/commits/d9c06c88c49eea99eaedc97b11abec25d6ba3650)). I'm not super satisfied with this fix, but it seems to reliable build and I haven't found anything else that did that. Do you mind taking a look and letting me know what you think?

I also updated a few dependencies that had gotten out of date ([899593c](https://github.com/FSHSchool/FSHOnline/pull/158/commits/899593ca76ed2e2383034ae3720a93e13ef4a1a8)), ran an `npm audit fix` ([3686de8](https://github.com/FSHSchool/FSHOnline/pull/158/commits/3686de8f0c7c8c75ed9b5be0f1b26f91fe317209)), and updated to the latest SUSHI ([8d42241](https://github.com/FSHSchool/FSHOnline/pull/158/commits/8d42241e82ccd1955c108f76f2b5823198895f34)).

cmoesel (approved):
Thank you, Julia, for this monumental effort. It looks great and works great. Good riddance, CRA!

mint-thompson (approved):
hooray for vite!

PR #159, opened 2024-06-10
Text editors support line wrapping

KaelynJefferson:
This PR includes:
- Functionality to display code with line wrapping in the code editors
- Checkbox for user selection of this functionality

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -216,7 +216,7 @@ export default function App(props) {
   const [isDragging, setIsDragging] = useState(false);
   const [configToShare, setConfigToShare] = useState({ canonical: '', version: '', fhirVersion: [], dependencies: '' });
   const [sharedConfig, setSharedConfig] = useState({});
-  const [textWrapped, setTextWrapped] = useState(false);
+  const [isLineWrapped, setIsLineWrapped] = useState(false);
 
   useEffect(() => {
     async function waitForFSH() {
@@ -314,8 +314,8 @@ export default function App(props) {
     setIsDragging(false);
   }
 
-  function setTextWrap(wrapSelected) {
-    setTextWrapped(wrapSelected);
+  function setLineWrap(wrapSelected) {
+    setIsLineWrapped(wrapSelected);
   }
 
   function debouncedMove(clientX) {
@@ -414,8 +414,8 @@ export default function App(props) {
             exampleMetadata={exampleFilePaths}
             isWaiting={isWaitingForFSHOutput || isWaitingForFHIROutput}
             saveAll={saveAll}
-            setTextWrap={setTextWrap}
-            textWrapped={textWrapped}
+            setIsLineWrapped={setLineWrap}
+            isLineWrapped={isLineWrapped}
           />
         </div>
         <div className={expandConsole ? classes.collapsedMain : classes.expandedMain}>
@@ -440,7 +440,7 @@ export default function App(props) {
                   isWaiting={isWaitingForFSHOutput}
                   setInitialText={setInitialText}
                   config={configToShare}
-                  textWrapped={textWrapped}
+                  isLineWrapped={isLineWrapped}
                 />
               </Grid>
               <Grid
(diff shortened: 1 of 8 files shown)


mint-thompson (approved):
Hooray for line wrapping!

src/App.js
```diff
@@ -216,6 +216,7 @@ export default function App(props) {
   const [isDragging, setIsDragging] = useState(false);
   const [configToShare, setConfigToShare] = useState({ canonical: '', version: '', fhirVersion: [], dependencies: '' });
   const [sharedConfig, setSharedConfig] = useState({});
+  const [textWrapped, setTextWrapped] = useState(false);
```
cmoesel, line 219: I noticed that in some places you call this text wrap and in other you call it line wrap. Unless there is a reason for the distinction, I think it might be good to call it the same thing everywhere (and I _think_ line wrap is more common).

Also, since it is a boolean, it might be more consistent with the conventions seen in the rest of this code to call it `isLineWrapped` and `setIsLineWrapped`.

src/components/FSHControls.js
```diff
@@ -116,6 +116,7 @@ export default function FSHControls(props) {
   const [fhirVersion, setFhirVersion] = useState('');
   const [dependencies, setDependencies] = useState('');
   const [isGoFSHIndented, setIsGoFSHIndented] = useState(false);
+  const [isGoFSHLineWrap, setIsGoFSHLineWrap] = useState(props.textWrapped);
```
cmoesel, line 119: Why is this called `isGoFSHLineWrap` and `setIsGoFSHLineWrap`? I think it applies to both sides of the app, so a more appropriate name might be `isEditorLineWrap` and `setEditorLineWrap` -- or even just `isLineWrap` and `setLineWrap`.

cmoesel (changes requested):
This feature works very well. I like it!

I noticed some inconsistencies in how we name things in the code. I think it might be easier for future maintainers if we are consistent in our naming.

src/tests/components/FSHControls.test.js
```diff
@@ -242,6 +242,52 @@ it('calls GoFSH with the indent option if the configuration checkbox is checked'
...
+    expect(isLineWrappedCheckbox).toBeChecked();
+    expect(resetLogMessages).toHaveBeenCalledTimes(1);
+    expect(runGoFSHSpy).toHaveBeenCalledWith([JSON.stringify(examplePatient, null, 2)], {
+      dependencies: [],
+      indent: false
+    }); // No IG resource added because canonical and version set to defaults
+    expect(onGoFSHClick).toHaveBeenCalledTimes(2);
+    expect(onGoFSHClick).toHaveBeenCalledWith('', true); // Loading
+    expect(onGoFSHClick).toHaveBeenCalledWith(simpleFsh, false);
+  });
```
jafeltra, line 288: I'm wondering about this test and what we really want to be testing. I don't think we need the checks on lines 280-287 because we don't really need to check the functions that get called as a result of trying to run GoFSH. That probably means we also don't need the actions on lines 273-276 to actually run GoFSH unless we're trying to use that to also check line wrapping in the GoFSH editor, but I don't think this test actually checks that right now.

That pretty much leaves the expect statement on 279 which checks if the checkbox is checked after we click it. I guess it is probably good to test that, so maybe it is worth keeping this test. It might be good to check that when you click the checkbox, that value makes it to the editor component, but that would require changing this test, and maybe not having the test in this file. I'm not sure if we have any tests that check the interaction between the components like that. If that gets complicated though, it might be worth stepping back and considering if that is worth testing.

jafeltra (changes requested):
I made a few comments inline. Let me know if you want to talk about any of them, especially the one about the test.

jafeltra (approved):
This looks good to me! The code and tests make sense and work as expected.

Since this configuration option is likely a personal preference that is pretty consistent, it might be nice to have the browser store this setting so it stays across refreshes and reopening the page. But that would be a separate task, if others even like it.

cmoesel (approved):
This works well! Thanks for the tests!

PR #162, opened 2024-06-21
Best Practices

jafeltra:
This PR adds the Best Practice document, a PR template, and updates the Contributing document with the latest updates.

This Best Practice document updates the testing framework bullet point to include the React Testing Library and removes the requirement to check coverage since we don't generate a coverage report in this repo. It also updates to use the `lint-fix` command that exists in this repo, and it turns out this repo never had a `prettier-fix` command, so I added one and documented the slight difference in command name. I also removed the requirement to update FSH School documentation since we don't document FSH Online there beyond linking it.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,5 @@
+**Description:**
+
+**Testing Instructions:**
+
+**Related Issue:**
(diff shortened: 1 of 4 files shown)


cmoesel (approved):
This looks good. Thank you!

PR #175, opened 2024-09-04
Various Dependency Updates

### LINKED ISSUE #163: Update Dependencies and Add DEPENDENCY-NOTES.md
Update dependencies to latest version where possible and document any dependencies that cannot be updated in DEPENDENCY-NOTES.md. Even if all dependencies can be updated, create a starter DEPENDENCY-NOTES.md file with the following contents:
```md
As of 2024 MMM DD:

The `npm outdated` command reports the following dependencies as outdated.
They are not being updated at this time for the reasons given below:
* **N/A**: No outdated dependencies
```

jafeltra:
**Description:** This PR updates various dependencies and adds a `DEPENDENCY-NOTES.md` file. There were some updates that I decided were out of scope for the current task, but should be revisited later based on priorities and time availability, which are documented in the file.

There were two updates that required updates to the code:

- Updating SUSHI to v3.11.1: When updating to this version of SUSHI, something in the Winston logger changed and it tries to log messages to stdout, and even though it should be using regular console.log logging when that isn't available in the browser, the messages weren't intercepted properly. To get around this, I defined a stub for `console._stdout.write` (which is what winston is using to log its messages) and had it use our same override function as the console.log messages.
- Updating `react-router-dom` to major version 6: This major update required changing how we access the URL path parts in `/share` and `/gist` links and changed how the `Route`s are set up.

**Testing Instructions:** Test that the app still runs in development and production. Because of the two updates above, also make sure that routing still works as expected and that the log messages show up in the FSH Online Console as expected when you run SUSHI and GoFSH.

**Related Issue:** Fixes #163


cmoesel (approved):
It all looks good to me!

mint-thompson (approved):
Tests are passing, and the application appears to be running normally. Thank you!

PR #176, opened 2024-09-05
Fix Example URLs

cmoesel:
**Description:** This PR fixes a bug that was causing examples with spaces in their path (e.g., `Code Systems`, `Rule Sets`, and `Value Sets` examples) to load as `404: Not Found`. The issue was that the constructed example URLs had a double-slash (`//`) where a single slash (`/`) should be -- which caused GitHub to issue an HTTP 301 redirect, but in doing so, it re-encoded the `%` in `%20`, resulting in an invalid URL.

I've also added a script to enable `npm run check` for consistency with other FSH repos.

**Testing Instructions:** Run FSH Online and confirm that you can load examples in `Code Systems`, `Rule Sets`, and `Value Sets`. Run `npm run check` and confirm it runs the tests, lint, and prettier.

**Related Issue:** N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
     "lint-fix": "eslint \"./src/**/*.{js,jsx}\" --fix",
     "prettier": "prettier --check \"**/*.{js,ts,jsx,tsx}\"",
     "prettier-fix": "prettier --write \"**/*.{js,ts,jsx,tsx}\"",
-    "test": "vitest"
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
Good find and fix!

mint-thompson (approved):
All the examples are working fine for me. Thank you!

PR #184, opened 2024-09-13
Move FSHOnline to https://fshonline.fshschool.org/

cmoesel:
**Description:** The FSHOnline and FSHOnline-Examples repos are moving to the GitHub FHIR organization. Due to how GitHub pages work, this necessitates moving FSH Online from https://fshschool.org/FSHOnline to https://fshonline.fshschool.org. This PR contains the changes necessary to serve the app at that subdomain, as well as changing links from github.com/FSHSchool/* to github.com/FHIR/*.

The PR should only be merged _after_ the repositories have been moved to the github.com/FHIR organization. In addition to merging this PR, the fshschool.org domain will need to be configured as described [here](https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/managing-a-custom-domain-for-your-github-pages-site#configuring-a-subdomain).
* Note since we are using a static generated site, you can skip step 4.
* In step 5, you should create a DNS `CNAME` record to point `fshonline.fshschool.org` to `fhir.github.io`.

**Testing Instructions:** The tests should pass and you should still be able to run this app locally. Note that now you test the app at the root of the URL, not at the /FSHOnline subpath. Also note that Examples will not work until _after_ the repos are moved -- so if you are testing _before_ the repos move, then you should not expect to see any content in the Examples modal.

**Related Issue:** N/A

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -122,14 +122,13 @@ export async function decodeFSH(encodedFSH) {
     return '';
   } else {
     const promisedURL = await expandLink(encodedFSH);
-
-    // Removes the encoded data from the end of the url, starting at index 40
-    const sliced64 = promisedURL.long_url.slice(40);
-    if (!promisedURL.long_url.includes('https://fshonline.fshschool.org/#/share/') || sliced64.length === 0) {
+    const encodedData = promisedURL.long_url?.match(
+      /^https?:\/\/(fshonline\.fshschool\.org|fshschool\.org\/FSHOnline)\/#\/share\/(.*)/
+    )?.[2];
+    if (!encodedData || encodedData === '') {
       return '';
     } else {
-      const displayText = inflateSync(Buffer.from(sliced64, 'base64')).toString('utf-8');
-      return displayText;
+      return inflateSync(Buffer.from(encodedData, 'base64')).toString('utf-8');
     }
   }
 }
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -63,7 +63,7 @@ const useStyles = makeStyles(() => ({
   }
 }));
 
-const githubURL = 'https://raw.githubusercontent.com/FSHSchool/FSHOnline-Examples/main';
+const githubURL = 'https://raw.githubusercontent.com/FHIR/FSHOnline-Examples/main';
 const defaultInfoMessage = 'There are no messages to display in the console.';
 const defaultProblemMessage = 'There are no problems to display in the console.';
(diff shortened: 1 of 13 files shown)


src/App.jsx
```diff
@@ -123,9 +123,9 @@ export async function decodeFSH(encodedFSH) {
   } else {
     const promisedURL = await expandLink(encodedFSH);
 
-    // Removes the encoded data from the end of the url, starting at index 38
+    // Removes the encoded data from the end of the url, starting at index 40
     const sliced64 = promisedURL.long_url.slice(40);
-    if (!promisedURL.long_url.includes('https://fshschool.org/FSHOnline/#/share/') || sliced64.length === 0) {
+    if (!promisedURL.long_url.includes('https://fshonline.fshschool.org/#/share/') || sliced64.length === 0) {
```
jafeltra, line 128: I have a question about this line. If someone tries to visit an old FSH Online share link, the `promisedURL` will be the returned results from Bitly. So for old share links that were made before the repos move, I think the `promisedURL.long_url` will be `https://fshschool.org/FSHOnline/#/share/...` when it gets here. Is there some redirect that happens somewhere within Bitly's resolution of the long URL it knows about that will let the `promisedURL.long_url` value to be udpated to be in the `fshonline.fshschool.org` format? If not, should we just allow either format here so that we properly decode the URL and load in the shared FSH?

Right now, when I tried to load an old share link, it doesn't work because the long URL is in the wrong format and the text doesn't load, which is why I'm wondering about a sneaky redirect that might not be happening yet.

jafeltra (commented):
I left one comment with a question about the Share functionality.

src/App.jsx
cmoesel: Oh. That's an excellent point. I think you're right. I was thinking (for some reason) that the bitly just had the encoded FSH. I forgot it was still essentially a URL. So we'll need to account for either URL in the code.

cmoesel: Fixed in fb0c1c2.

cmoesel:
Thanks for the review, @jafeltra. I've updated the share URL decoding in fb0c1c2 so it is more robust now - and handles both the old and new links. In addition, I threw in a few tests to confirm the behavior.

jafeltra (approved):
Thanks for the update and the tests! This looks good to me!

mint-thompson (approved):
Tests look good, documentation updates look good, and everything works when run on my system. Thank you!

PR #188, opened 2024-09-25
Avoid using an object when displaying id in JSON editor sidebar

jafeltra:
**Description:**
This PR avoids returning an object as the `id` value that is used when we display FHIR artifacts in the JSON editor sidebar.

When an `id` is set using an instance, the `id` property is an object, not a string. However, FSH Online assumed `id` was always a string value and passed it to components assuming it was a string.

To handle this, I updated the `checkIdToDisplay` function that we use to determine an `id` value to better handle objects.

**Testing Instructions:**
The main goal of this PR was just to prevent FSH Online from crashing. You can test this by using the following FSH:

```
Instance: MyId
InstanceOf: string
Usage: #inline
* value = "my-id"

Instance: MyObs
InstanceOf: Observation
* id = MyId
* code = #123
* status = #final
```
On master (and the deployed FSH Online), running SUSHI on this FSH will cause FSH Online to crash. On this branch, it should work and use `my-id` as the value in the sidebar.

The goal of this wasn't to make SUSHI work any better, so there is still an error you'll see coming from SUSHI, and if you run SUSHI in a terminal, the files it creates is probably named wrong, but I just wanted to avoid a full crash of the website.

**Related Issue:** N/A

### DIFF
--- a/src/components/JSONEditor.jsx
+++ b/src/components/JSONEditor.jsx
@@ -116,7 +116,15 @@ const checkFshType = (def) => {
 
 const checkIdToDisplay = (def) => {
   if (def.id) {
-    return def.id;
+    if (typeof def.id === 'object') {
+      // NOTE: This case shouldn't really happen -- it only comes up when id was assigned using an inline instance
+      if (def.id.value) {
+        return def.id.value;
+      }
+    } else {
+      // Typical case
+      return def.id;
+    }
   }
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
As advertised!

mint-thompson (approved):
hooray for not crashing!

PR #193, opened 2024-11-08
Remove Bitly API Use

jafeltra:
**Description:**

This PR removes our use of the Bitly API. In order to no longer tie our share functionality to an individual team member, we are going to remove the use of Bitly to shorten the link for sharing FSH. The functions that used the Bitly API are still in the code in case this functionality is wanted in the future. The functions should work (as long as the API still works the same) if anyone wants to use them in the future.

Now that there are no longer short links, the full-length long URL is now supported when loading in FSH snippets. The modal that opens when you go to share your FSH now scrolls properly so the box stays a reasonable size when the link is long. The "Copy" button now copies a formatted link, which is nice so that users aren't pasting really long URLs that look kind of ugly everywhere. If you copy the link, you should see it formatted properly as markdown in Zulip and as a hyperlink in places like Word.

In order to preserve existing short shared links, I grabbed all previously created links and added them as a file within `public/shares/`. Each file name is the short URL and the file content is the expanded URL. This should make it so that any share link that is out in the wild will still continue to work.

I removed the `VITE_BITLY_KEY` environment variable from the deploy script. I can also remove it from GitHub once this is ready to be merged.

**Testing Instructions:**

- Try creating a new share link and paste the formatted URL in Zulip and in other places, like Word.
  - I decided that if you didn't want a formatted version of the link, you could just copy the displayed link manually. However, if you think having an "unformatted copy" button would be helpful, let me know.
- Try loading a long link.
- Try loading on a previously shared (short) link.
- Take a look at the tests to make sure they cover the updated functionality.
- If you want, you can also build the app (`npm run build`) and try using the production build (`npm run preview`) to make sure the built app doesn't seem to slow down with all the new files in `public`.

There were lots of approaches that I tried for supporting existing short links, so if you don't like the approach with `public/shares/`, let me know and I'm happy to chat.

(Sorry that you have to scroll past 1645 files to get to the majority of the files that changed).

**Related Issue:**
N/A


mint-thompson (commented):
I think this is a good strategy for keeping the old links working while removing bitly. There are a lot of files, but they're not big files, so that seems fine to me. One thing I want to report is: when I tried running this locally, the "copy link" button didn't seem to work for me (Chrome, Windows). My clipboard would end up empty. I noticed that the ordering of components in `ShareLink.jsx` changed: was that intentional?

jafeltra:
Great catch @mint-thompson. I updated what gets copied on the clipboard and the format in [0816778](https://github.com/FHIR/FSHOnline/pull/193/commits/081677858ff24a9008a587de9825a878d7072cbb). This changes it so that it copies in markdown format, so it should paste in plain text documents (like in VS Code / another editor). When you paste into Zulip, I was also seeing it paste as markdown format (and not get wrapped in ``s as code). This change does make it so that it isn't pasted into places like Word/email as a hyperlink, but I decided that that might not be as important as the other two case. Let me know if anyone disagrees with that. If we want to support the hyperlink format in Word/email, we might just need to add a second button for a "Copy as Hyperlink".

jafeltra:
@mint-thompson - [989a0df](https://github.com/FHIR/FSHOnline/pull/193/commits/989a0df8043f7ef06fe72444ed05be4085fdd54e) adds a separate button for the markdown formatted copy button and the regular copy. I had the tooltip for that button say "Copy Formatted Link to Clipboard" and used the "Code" icon for it. I was debating having the tooltip say "Copy Markdown Formatted Link" or "Copy Formatted Link for Zulip" or something those lines, but decided to keep it simpler. Let me know what you think of the wording and the icon choice.

mint-thompson (commented):
The updated copy buttons work perfectly! As for the button tooltips, maybe they could start as "Copy Link to Clipboard" and "Copy Formatted Link to Clipboard", then change to "Link Copied" and "Formatted Link Copied" when clicked.

As a minor style thing to look at, there's some extra space to the right of the copy buttons. If the box containing the text were wider and the box containing the copy buttons were narrower, there wouldn't be that extra space. It's not a big deal if it ends up being resistant to change, though.

image

jafeltra:
Hm that is not the styling I was intending, so I'll take a look!

public/shares/fakeurl
cmoesel, line 1: Is `fakeurl` real? Do we need it? I tried loading it and it didn't load any content.

src/AppRouter.jsx
```diff
@@ -5,7 +5,8 @@ import App from './App';
 export default function AppRouter() {
   return (
     <Routes>
-      <Route path="/share/:id" element={<App path="share" />} />
+      {/* use a splat for share because compressed URL string may have / in it */}
+      <Route path="/share/*" element={<App path="share" />} />
```
cmoesel, line 9: Oooh. Good catch!

public/shares/fakeurl
cmoesel: Oh. Nevermind. I see now that this is used in testing?

cmoesel (dismissed):
I think we have finally solved it, @jafeltra. Good work!

public/shares/fakeurl
jafeltra: Ohh you know what, I had been using it for testing, but I just realized that now that I mocked out the response in the test, I don't actually need the file. I'll remove it.

jafeltra: I removed it in [fb322e6](https://github.com/FHIR/FSHOnline/pull/193/commits/fb322e620d1d8fc6b8c1926701b8b66802e09c70), which means I also dismissed your approval.

cmoesel (approved):
Thank you, Julia!

PR #201, opened 2025-03-28
Color level of log message

### LINKED ISSUE #180: Colorized text in the Console
Transferred from internal JIRA issue originally created 2020-07-30.

While the console is functional, it would be more user-friendly if the log messages in the Console pane were colored based on their error/warning/info status. This would be more similar to how the messages appear in a terminal when you run SUSHI/GoFSH. Typically, the "error"/"warn"/"info" words are colorized and the box with the SUSHI/GoFSH results is colorized.

jafeltra:
**Description:**
This is a small PR to add a block of color over the log level for messages in the Console.

Note that the "results" box isn't colored like it is for SUSHI and GoFSH in the terminal.

**Testing Instructions:**
Try out FSH Online and view different messages at different log levels. Make sure to view an info, error, warning, and debug level message.

**Related Issue:** 
Fixes #180 (with the exception of coloring the results box).

### DIFF
--- a/src/components/Console.jsx
+++ b/src/components/Console.jsx
@@ -20,25 +20,18 @@ const useStyles = makeStyles((theme) => ({
     overflow: 'scroll',
     boxSizing: 'border-box'
   },
-  button: {
-    marginRight: '10px',
-    color: theme.palette.common.white,
-    textTransform: 'none',
-    padding: 0
-  },
-  problemsButton: {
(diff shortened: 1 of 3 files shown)


cmoesel (approved):
This is great. It makes it much easier to pick out the errors and warnings in the console logs. Thanks for this, @jafeltra!

